PR #68, opened 2020-07-07
Rework benchmarking system

felixhekhorn:


AleCandido:
## Towards a stable Benchmarks
We are currently refactoring the whole benchmarking system (not regression test) to make it stable and suitable for CI workflows

The following steps are in our plan:

### Redefine the db structure:
- **apfel** will have its own database, with its theories, observables and cache
- **qcdnum** as well, both theories and observables won't be the same of apfel, because of the different implemented features in the two codes
- **logs** will have their own db, used mainly locally (even if it will be produced also by the workflows)
- **sandbox** input db will be added to support the new structure

#### observables
at the same time we are going to differentiate observables per external program, but also unifying them in few records (essentially one per perturbative order), to be able to speed up the calculation either in the external objects (that will precompute everything on a grid) and in the same `yadism`, that will take advantage of caching system and numba JIT compilation

### Less dependencies into the workflows
- the cache will be uploaded in the same way of what it is happening for regression, since it is even more stable than regression itself
- apfel computation won't be performed anymore during workflows
- qcdnum comparison will be available for workflows
- conda won't be anymore a part of the game

### Welcome sandbox
- a new way of running comparisons will be available for local use only
- it will have its dedicated script ecosystem, to be able to change comparison in a fast and flexible way
- it won't affect any test/benchmark score so there is no reason to worry about its stability

AleCandido:
- [ ] think about list of known differences for make benchmarks aware of particular kinematical corners

Issue #61, opened 2020-06-02
More benchmarks


AleCandido:
- [x] Include benchmarks suggested by Juan Rojo, [see the wiki](https://github.com/N3PDF/yadism/wiki/Benchmarks-resources).
- [x] Include even more benchmarks, for example the code by Paolo Nason/Eric Laenen we discussed with Stefano C. (you can find it in https://github.com/nnpdf/external/tree/master/FONLLdis)
- [x] `xspace-bench`

AleCandido:
Working on QCDNUM, still to be done:

- [ ] include & install QCDNUM python package
- [ ] upgrade navigator
- [ ] compare QCDNUM and find out numeric issues
- [ ] extend QCDNUM bindings (heavy flavors, and so on)

felixhekhorn:
We give up on FONLLdis because it is too old, complicated and limited

AleCandido:
Spurious minus sign surviving in F3bottom:
https://github.com/N3PDF/yadism/blob/develop/benchmarks/yadmark/benchmark/external/xspace_bench_utils.py#L197-L198

felixhekhorn:
finished

PR #91, opened 2020-12-02
xspace-bench benchmarks

felixhekhorn:
completes ~~FONLLdis~~ `xspace-bench` in #61


giacomomagni:
Xspace-bench is now working for F2, F3, FL total below bottom threshold.

Issue #82, opened 2020-10-15
Intrinsic charm


felixhekhorn:
Literature:
- PhysLett.: https://inspirehep.net/literature/1395621
- EPJC: https://inspirehep.net/literature/1464831
- JHEP: https://inspirehep.net/literature/1397009

Issues:
- remember to update the weights for `IC=0`, current scenario is:
  - consistent `IC=0` for NC at the level of the coefficient functions (LO not provided for heavy structure functions), not sure at the level of the weights (maybe they are provided also for heavy inputs)
  - consistent `IC=1` (but also `IB=1` and `IT=1`) for CC: all the weights are always implemented and coefficient functions as well, for both light and heavy inputs
- the correct setup for `IntrinsicQuark` (=`IQ`) specified in `[m,n)` range should be:
  - the weights should be implemented up to `n` (excluded), because all the heavy not intrinsic inputs should not be present as PDFs
  - the `FNS` scheme should be consistent, so only the flavors up to `m` (excluded) can be considered as light in the fitting region (than if it is a `VFNS` it can change at higher `Q2`)

Note that the heavy object here is referred to the input, not to the output/object coupling to the vector, so the heavy structure functions should be always implemented for both heavy and light inputs, and are the weights that controls if they are needed or not, not the coefficient functions (so the fact that is actually working for the NC is just a chance)

PR #93, opened 2020-12-22
FONLL for intrinsic

felixhekhorn:


felixhekhorn:
will resolve part of #82

PR #95, opened 2021-01-26
New benchmark runners

giacomomagni:
Benchmarking using banana.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       max-parallel: 3
       matrix:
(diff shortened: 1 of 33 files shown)


AleCandido:
Things to be fixed:
- xspace_bench
  - [ ] F3total CC: gonly || 6 < Q2 <16 (since there is a cancellation for the total below 6)
  - [ ] LAMBDA in Fortran code (if possible remove from Fortran, do in python, otherwise comment it)
  - [ ] F3light NC donly/dbaronly not working
  - [ ] F3 CC has the wrong sign according to projectile: check usage of projectile
- qcdnum
  - [ ] Fbottom: 22 < Q2 < 40
  - [ ] ZM not matching

PR #96, opened 2021-01-29
More unittests

felixhekhorn:


tests/input/test_defaults.py
```diff
@@ -0,0 +1,68 @@
...
+
+import yaml 
+
+from yadism.input import errors, defaults
+
+repo_path = pathlib.Path(__file__).absolute().parents[1]
+
+
+# @pytest.mark.skip
+class TestDefaultManager:
```
AleCandido, line 17: `DefaultManager` it's legacy code: I don't know if it will ever be used, but for sure not in the near future, and maybe we should immediately trash it

PR #99, opened 2021-02-04
New benchmark runners

AleCandido:
#95

### DIFF
--- a/benchmarks/runners/qcdnum_bench.py
+++ b/benchmarks/runners/qcdnum_bench.py
@@ -1,12 +1,6 @@
 # -*- coding: utf-8 -*-
 #
 # Compare the results with QCDNUM
-
-# import pytest
-# from yadmark.benchmark.db_interface import DBInterface, QueryFieldsEqual
-
-import pathlib
-import copy
 import pytest
 import numpy as np
 
@@ -25,7 +19,6 @@ class QCDNUMBenchmark(Runner):
 
 
 class BenchmarkPlain(QCDNUMBenchmark):
-
     """The most basic checks"""
 
     def benchmark_lo(self):
@@ -41,7 +34,6 @@ class BenchmarkPlain(QCDNUMBenchmark):
 
 @pytest.mark.skip
 class BenchmarkScaleVariations(QCDNUMBenchmark):
-
     """Vary factorization and renormalization scale"""
 
     @staticmethod
@@ -70,8 +62,8 @@ class BenchmarkScaleVariations(QCDNUMBenchmark):
     @staticmethod
     def theory_updates(pto, FNS):
         # There is a QCDNUM error: "STOP ZMSTFUN: You cant vary both Q2 and muR2 scales --> STOP"
-        # this is a limitation of QCDNUM in principle, so you have to work around it, i.e. fix Q2 and only
-        # vary muR or vice versa
+        # this is a limitation of QCDNUM in principle, so you have to work around it, i.e. fix Q2
+        # and only vary muR or vice versa
 
         if FNS == 1:
             sv = {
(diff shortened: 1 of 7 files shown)


giacomomagni:
From my side it's possible to merge. I added a small modification in the yadmark runner to make it more flexible with the order of the points returned by the external program.

felixhekhorn:
running with `pytest` I still get 2 errors for `xspace-bench`:

```
E                   ValueError: Sort problem: x and/or Q2 do not match.
...
======================================== short test summary info ========================================
FAILED benchmarks/runners/xspace_bench_bench.py::BenchmarkPlain::benchmark_lo - ValueError: Sort probl...
FAILED benchmarks/runners/xspace_bench_bench.py::BenchmarkPlain::benchmark_nlo - ValueError: Sort prob...
```

please make sure to have pytest report everything passed

felixhekhorn:
and in `qcdnum` I get
```
E                   NotImplementedError: flavor total is not implemented!
```

benchmarks/runners/qcdnum_bench.py
```diff
@@ -27,7 +27,7 @@ def benchmark_lo(self):
     def benchmark_nlo(self):
         self.run(
             [{"PTO": 1}],
-            observables.build(**(observables.default_config[1])),
+            observables.build(**(observables.default_config[2])),
```
AleCandido, line 30: We're using `observables.default_config` to store the default configuration for any perturbative order, so the number it's not just a random label, but it's meant to be PTO.

The idea of saving other configurations it's not bad, simply make a new dictionary in `observables` module, and try to make a sensible choice (a general one you can reuse more than once, and then update the few things that are different).

benchmarks/runners/xspace_bench_bench.py
```diff
@@ -31,7 +31,7 @@ def benchmark_nlo(self):
             ["ToyLHAPDF"],
         )
 
-
+@pytest.mark.skip
```
AleCandido, line 34: `pytest` should always run all the provided benchmarks, previous occurences of `skip` where just temporary, because we did not have the time of making them work properly. Since you are exactly making them work, remove any `skip` before merging and test everything you wrote.

benchmarks/yadmark/data/observables.py
```diff
@@ -31,6 +31,10 @@
         "observable_names": ["F2light", "F2total", "FLtotal", "F3total"],
         "kinematics": default_kinematics,
     },
+    2: {
```
AleCandido, line 34: See the other comment, simply split this config out of this dictionary. Or in this case maybe it's better to simply update the `default_config[1]`, since the two are so close to be almost the same.

felixhekhorn (changes requested):

benchmarks/yadmark/data/observables.py
```diff
@@ -28,10 +28,6 @@
 default_config = {
     0: {"observable_names": ["F2light"], "kinematics": default_kinematics},
     1: {
-        "observable_names": ["F2light", "F2total", "FLtotal", "F3total"],
-        "kinematics": default_kinematics,
-    },
-    2: {
         "observable_names": ["F2light", "FLlight", "F3light"],
```
felixhekhorn, line 31: This is not a good idea - the default should still be total, since this is available for everybody in contrast to light

felixhekhorn (changes requested):

benchmarks/runners/qcdnum_bench.py
```diff
@@ -26,9 +26,14 @@ def benchmark_lo(self):
         self.run([{}], observables.build(**(observables.default_config[0])), ["ToyLH"])
 
     def benchmark_nlo(self):
+
+        fnames =  {"observable_names": ["F2light", "FLlight", "F3light"],}
+        obs =  observables.default_config[1]
+        obs.update(fnames)
```
AleCandido, line 32: Pay attention to this: `dict` are mutable objects, so if you change one of them and then use again you will keep using the updated one.
I would suggest to add a `.copy()` call:
```py
        obs =  observables.default_config[1].copy()
        obs.update(fnames)
```

felixhekhorn:
Remember to mark the conversations as "resolved" *if* it is ;-)

benchmarks/runners/qcdnum_bench.py
giacomomagni: okay done

PR #112, opened 2021-05-14
Parallel

AleCandido:
1. Computing 1 ESF is not expensive for `yadism`
2. However it scales linearly with the number of points
3. Different ESFs for the same observable have no interactions

Then it's easy to see that if the number of requested points it's small there is no performance issue to deal with and if the number of requested ESFs is high there is an easy solution: **parallelize the computation of ESFs for a given observable**.

### To Do
- [ ] add non-parallel mode (it is a good idea to avoid if not appreciated...)
- [ ] refactor the mess in `Runner.get_result()`

#### Note
Different observables might interact (like in the case of TMC or cross-sections) because of cache (that is on its own a performance improvement), so it is not a good idea to parallelize across different ones.


AleCandido:
Actually on my 6 x `Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz` (x 2 for hyperthreading) I got a speedup by a **factor of 3**.

AleCandido:
Performances of numbified coefficient functions are extremely good, than there is no need of extra complications.

PR #113, opened 2021-05-18
Scale variations

felixhekhorn:


AleCandido:
- [ ] separate nonsinglet, singlet, gluon kernels&operators in order to apply scale variations (they apply differently to each of them)
- [ ] deny factorization scale variations (but not renormalization ones) for intrinsic (and so **maybe** matching)
- [ ] think about recombining s, ns, g (e.g. gVV and gAA, or FONLL bits) before or after blow-up (i.e. blow-up before or after splitting functions)

PR #115, opened 2021-07-06
Fix SV @ NNLO

felixhekhorn:


felixhekhorn:
We have at the moment 3 failing components:
https://github.com/N3PDF/yadism/blob/61aef7a172b58d91e9797dc5dffc6f7bc240258c/tests/cf/test_splitting_functions.py#L95
https://github.com/N3PDF/yadism/blob/61aef7a172b58d91e9797dc5dffc6f7bc240258c/tests/cf/test_splitting_functions.py#L104
https://github.com/N3PDF/yadism/blob/61aef7a172b58d91e9797dc5dffc6f7bc240258c/tests/cf/test_splitting_functions.py#L118

PR #116, opened 2021-07-15
Introduce poetry

AleCandido:
Add:

- [x] poetry
- [x] poetry dynamic versioning
- [x] poetry task runner
- workflows
  - [x] release
  - [x] test


AleCandido:
I also removed old and unused stuffs, and the only two pain points left behind are:
- non-pip dependencies (that have to be installed manually into the poetry environment)
  - and therefore the problem of running benchmarks in CI
- incomplete coverage (i.e. unit tests) that was not the goal of this PR

benchmarks/pyproject.toml
```diff
@@ -0,0 +1,27 @@
+[build-system]
+requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning"]
+build-backend = "poetry.core.masonry.api"
+
+[tool.poetry]
+name = "yadmark"
+version = "0.0.0"
+description = "yadism benchmarks"
+authors = ["Alessandro Candido <candido.ale@gmail.com>"]
```
felixhekhorn, line 9: Maybe you want to list the three of us? ;-)

src/yadism/__init__.py
```diff
@@ -3,7 +3,8 @@
 .. todo::
     docs
 """
-from .output import Output
+__version__ = "0.0.0"
```
felixhekhorn, line 6: this is also used inside the docs, isn't it?

felixhekhorn (changes requested):

felixhekhorn (approved):

PR #133, opened 2022-02-18
Provide tar output

### LINKED ISSUE #123: Output format revamped
Since the success in `eko` new output format implementation n3pdf/eko#76 I propose to follow the same path here, and so to split the operator from the metadata.
Main reason supporting this is that we'll save a lot in loading and dumping time.

### Output structure
Here the typical structure of output is a bit different from those in `eko`, and a bit more complicate (but still perfectly suitable).

An `Output` object is still basically a dictionary, e.g. with the following keys:

```py
[ins] In [19]: out.keys()
Out[19]: dict_keys(['XSHERACC', 'interpolation_is_log', 'interpolation_polynomial_degree', 'interpolation_xgrid', 'pids', 'projectilePID'])
```
At the moment we need an input runcard in order to be able to load the observable, because we don't know otherwise which are the observables (e.g. `XSHERACC`) names in the output object.

There are two possible solutions:
- we scope observables within a single entry `obsvervables` in the object instance
- we keep as it is, but we add an `observables` entry with a list of names of observables

While every other value is just a scalar (at most a list/array of scalars), each observable is a very nested object.

### Observables -> ESF structure

Each observable is a list of ESF, potentially of different length. In turn, each ESF has the same structure.

```py
[ins] In [14]: e = out["XSHERACC"][0]

[ins] In [15]: len(out["XSHERACC"])
Out[15]: 42

[ins] In [16]: np.array(e.get_raw()["orders"][0]["values"]).shape
Out[16]: (14, 50)

[ins] In [17]: e.get_raw()["orders"][0].keys()
Out[17]: dict_keys(['order', 'values', 'errors'])

[ins] In [18]: e.get_raw().keys()
Out[18]: dict_keys(['x', 'Q2', 'nf', 'orders'])
```

Thus we can make it an array, what we need is:
- one array per observable
- first dimension runs over ESFs (so we need to store `x`, `Q2`, and `nf` somewhere else, i..e in metadata)
- second dimension runs over orders; they are the same for each observable -> we need to store the list of orders
- two other dimensions run over flavor and interpolation grid (both already stored, respectively as `pids` and `interpolation_xgrid`)
- optionally one more dimension can run over `[value, error]`

## Recap
We need:
- a tar archive containing
- metadata, i.e. all key-value pairs, but observables, to which we should add
  - the list of orders
  - `x`, `Q2`, and `nf` for each ESF in each observable (so one list per observable, made of 3-tuples)
- dump one array per observable; it's to dump mandatory separately because they might not be uniform

AleCandido:
This should close #123 

- [x] dump tar
- [ ] load tar
- [ ] dump runcards as well (they can be kept separate in the tar)

Nice achievements:
- at the moment, on a simple output (LO only, 3 observables `FX_light`) we obtain a factor of roughly 60: from 1.9MiB to 32KiB (while 32KiB I believe something like "a few units of blocks", even an empty file is 4KiB)
- the saving speed is incredibly faster

Still I have to reload it, but looking inside at the moment it seems alright.

#### Considerations
It is nice to maintain an output independent of `pineappl`, since we can store all the information we want, including multiple observables (that makes sense if they are computed in a single run).
Of course, it's always possible to dump multiple PineAPPL grids, but in this way we have a one-to-one correspondance between runcards, runs, and outputs.


AleCandido:
P.S.: it is based on `feature/FONLL-B`, but I target `develop` the same, since I believe we'll merge FONLL-B soon

AleCandido:
@felixhekhorn I'll need your opinion on the **Memory Usage** section. Then, we decide what we want to do and what not, and I'll open a dedicated issue.

It's rather unrelated to this PR, but it came natural the moment I started dumping the objects (since the size of the objects on disk is so tiny).

AleCandido:
In [3931e9c](https://github.com/N3PDF/yadism/pull/133/commits/3931e9cf39313080cfaa4b67b823aed46ba95fa0) I solved a bug of the `yaml` output, it was loading everything as `ESFResult`, always discarding the information on `y`.

Now it's using `EXSResult` when relevant.

AleCandido:
In [4e00dfc](https://github.com/N3PDF/yadism/pull/133/commits/4e00dfc8f5409e8f96328ad19389e539f814955e) I reduced the size of the `yaml` by a factor 2-3, and improved its readability, just switching to `default_flow_style=None`.

AleCandido:
The new output format is ready (for what I can tell).

I'll write the docs by myself (the docstrings for sure, and I'll have a look if we have a section about IO, otherwise considering addition). I'd like instead some help in testing, both manually and the unit tests @felixhekhorn.

(@andreab1997 if you'll have spare time, you could in place of @felixhekhorn, who might be performing a task that is blocking yours, at some point)

AleCandido:
All done but testing.

andreab1997:
I added a simple test for dumping and loading tar outputs (which was the missing one if I've understood correctly). Do we need to test something else? @AleCandido @felixhekhorn @giacomomagni

tests/test_output.py
```diff
@@ -110,3 +110,31 @@ def test_io(self):
...
                 o3["interpolation_xgrid"], d["interpolation_xgrid"]
             )
+    
+    def test_tar_output(self, tmp_path):
+        out, obs = self.fake_output()
+        #dump fake output
+        o1 = output.Output(out)
+        temp_path = pathlib.Path(tmp_path)
+        fn = pathlib.Path("test.tar")
+        actual_path = pathlib.PurePath(temp_path, fn)
```
AleCandido, line 120: Why do you need a `PurePath`?

AleCandido: Even more: since `fn` you're using only once, you can just do:
```py
actual_path = temp_path / "test.tar"
```
and that's it :)

andreab1997: Yes, you're right. I fixed it.

PR #134, opened 2022-03-03
Drop TMC cache

AleCandido:
`yadism` is usually rather light as a program, but it keeps the whole cache of the full execution in memory, releasing it only on exit.

In most cases, the cache coincide with the result, and since the result is generally small, even the memory used won't be so large.

This is not true when Target Mass Corrections are enabled, since the final observable is the result of an intermediate integration over another layer of observables.
All the results are cached, and this is useful to speed up successive integrations (since the evaluation is always done on the same points), but it is actually the source of large memory footprint, since this cache is never released.

Since the cache is only useful for TMC observables evaluated at the same `Q2`, if the observables are sorted in `Q2` it can be dropped as soon as `Q2` changes.
This will lower the memory footprint to a negligible value, for all currently known use cases.

Preparation:

- [x] set up memory measurement (inside *airspeed velocity*) for:
  -  non-TMC calculation
  - TMC calculation

Solution:

- [x] add a mechanism for SF to drop the ~old ESFs~
  - ~this is only needed for SF, since XS are always `final` (in other words, if XS TMC is requested, it will be composed of SF TMC, not of non-TMC XS)~
- ~add `final` attribute to prevent cache to be dropped~
- [x] cache the current value of `Q2` for each SF
- [x] whenever requested a new `Q2`, drop the whole cache
- ~try to detect when an SF will be no longer used, and drop the full cache~
  - ~since there won't be a call for a different `Q2` in this case~
  - ~it can be simply done when the computation for an observable is over, calling the cache dropping mechanism on all the children and propagate (e.g.: XSTMC -> SFTMC -> SF)~
- [x] sort input kinematics on `(Q2, y, x)` (store the inverse mapping)
- [x] restore input sorting after computation

~Pay attention always **not to drop the results** (since the only way to distinguish result from cache, is that they belong to the outermost observable).
Most likely, the easiest way is to tag outermost SF and XS with something like a `final=True` attribute, to prevent the cache to be dropped.~
It looks like we saved cache and results in two different attributes:
https://github.com/N3PDF/yadism/blob/d8b9fbf40c2489379aa28c14bfe60475d46747d7/src/yadism/sf.py#L35-L36


AleCandido:
@felixhekhorn actually I believe that integration might be shared only if `x` is different, so the cache should be dropped also if `y` is changing. Is it correct?

felixhekhorn:
> @felixhekhorn actually I believe that integration might be shared only if `x` is different, so the cache should be dropped also if `y` is changing. Is it correct?

Mmm I'm not sure that is relevant: the cache is inside SF and structure functions (ESF) do *not* depend on y - y only enters on EXS, but that's not cached, correct? so SF does not know if and when y changes ...

AleCandido:
So, if `y` changes I should not drop the cache?

felixhekhorn:
> So, if `y` changes I should not drop the cache, is it correct?

in principle, I'd say yes, but how can you know inside SF?

AleCandido:
> in principle, I'd say yes, but how can you know inside SF?

I sort the observables such that they iterate faster on `y` than `Q2`, I cache the current value of `Q2` in `SF`, and I drop only when `Q2` requested from outside is changing.

felixhekhorn:
> Unfortunately, I'll need to sort twice, since I want to restore in the output the same sorting that has been passed in the input by the user.

and that you have to (as we already experienced) - but can't you just find the permutation?

AleCandido:
> and that you have to (as we already experienced) - but can't you just find the permutation?

Unfortunately, I believe that `sorted()` won't give me the permutation, but maybe `np.unique` (or something like, without dropping repeated values) yes.
I have to investigate a bit.

pyproject.toml
```diff
@@ -73,6 +73,7 @@ pytest-cov = "^2.12.1"
 pytest-env = "^0.6.2"
 # benchmark
 asv = "^0.4.2"
+virtualenv = "^20.13.2"
```
felixhekhorn, line 76: hu? why do we have an explict dependency to virtualenv (inside poetry)?

AleCandido: Because of *airspeed velocity* (that of course I want to run inside `poetry`).

I know there are plenty of things to be done, but I need to measure the improvement in memory footprint, so my idea was to set up a couple of quick benchmarks:
- compute non-TMC observable: nothing should change
- compute TMC observable: it should pass from a much larger memory consumption than the former, to a comparable one

felixhekhorn: I see, so when it measures it needs an env ...

AleCandido:
I reverted, and instead of having a complicated generator to which passing all sorts of hooks, I decided to make a uniform interface to the elements to iterate: `SF.elements` and `XS.elements`.
The result was stored anyhow at `ESF` level, so this really changes nothing, and now we have a much cleaner sorting and cache dropping.

https://github.com/N3PDF/yadism/blob/41b7dc85007c23ee1d737dfa70a2aa721e0b44dd/src/yadism/runner.py#L247-L267

AleCandido:
@felixhekhorn now we only miss testing and benchmarking :)z

AleCandido:
I improved slightly the coverage of `yadism`, sometimes not even doing a proper job (but something that would be easier to improve).

The actual memory footprint has been benchmarked, and for 1000 TMC ESF with different Q2 (roughly to 330 cross-sections with unique Q2 values, realistic scenario) the impact is actually sizeable (on 50 points `xgrid`).

image

If TMC is not active, the impact is almost negligible.

image

Everything as expected.

The only new discovery is that 250-300 MB are needed to run in any case, being Python and Numba somehow responsible for them.

(The spike in the plots is not understood, belonging to an almost irrelevant commit, but it's also irrelevant for the sake of this PR)

AleCandido:
For me, we can merge.

src/yadmark/data/observables.py
```diff
@@ -20,7 +20,7 @@
 
 default_kinematics = []
 default_kinematics.extend(
-    [dict(x=x, Q2=10.0, y=0) for x in default_card["interpolation_xgrid"][3::3]]
+    [dict(x=x, Q2=10.0, y=0) for x in default_card["interpolation_xgrid"][12::3]]
```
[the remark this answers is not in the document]
AleCandido: Because otherwise APFEL is crashing: it cries that `x` is too low

felixhekhorn (approved):
I fixed two bugs - apart from that seems alright: tests are running (if only we had more :innocent: ) and benchmark are also still running ...

PR #138, opened 2022-06-01
Init NNLO CC heavy quark

AleCandido:
Include Jun Gao grids in yadism


AleCandido:
Now we might have a problem: Jun Gao's grids are not uniform in `(z, x, Q2)` space...

Following the plot of xgrid elements per position (in the xgrid) and block (a unit in Jun Gao's files) for a single file (i.e. one of $\nu$ or $\bar{\nu}$, and one value of $y_{Bj}$).
nbxsec1 00 0m1 51y1-xgrid

Next, a plot of relevant kinematics ($x_{Bj}$ and two mass scales, still to understand what they are exactly):
nbxsec1 00 0m1 51y1-xbj-scales
(This last plot display only the trend, to have a more meaningful one I normalized and took logs, the actual ranges are written as a title)

AleCandido:
What do we want to do now? @felixhekhorn 

The `xgrid` (meaning `zgrid`, i.e. the PDF variable, first plot) is quite severe, we are really missing part of the data.

The other thing I don't know is if this is the pure NNLO correction (I hope) or the whole NNLO value.
In the first case we can simply extend the grid with zeros, and that's it (not very pleasant, but acceptable if you take into account it was considered as 0 before), in the second fill with NLO (that is effectively the same solution, but this case would be very bad...).

felixhekhorn:
Mmm, I'm not sure I understand ...
- first of all, we can always ask Jun for some explanation
- the plots should show the usual statement that the integration range (dz) is bound between some lower limit related to x and an upper limit (if i remember in CC was 1, maybe?)
- to determine whether this the data is sufficient you have to compare to e.g. Chorus data
- also the program might well be made for this so notation might be similar (e.g. E)
- I'd also expect the grid to contain the full (all-order) calculation (as said might be tailored for Chorus)

giacomomagni:
What is the status of this PR? Shall we convert to issue?

AleCandido:
> What is the status of this PR? Shall we convert to issue?

I would trim down to just `extras/fonll-cc-heavy-quark`, but keep that part, since it contains the Jun Gao parser (at least I won't have to work it out again...).

felixhekhorn:
I suggest to merge now as is* and then to start a new PR

PS: *why is `poetry.lock` changed?

AleCandido:
I will drop even the changes to the tests

PR #139, opened 2022-06-07
Expose XS computation

AleCandido:
Expose the coefficients computation, such that it can be used even outside yadism itself.

### DIFF
--- a/src/yadism/esf/exs.py
+++ b/src/yadism/esf/exs.py
@@ -27,11 +27,6 @@ def xs_coeffs(kind, y, x=None, Q2=None, params=None):
     ym = 1.0 - (1.0 - y) ** 2
     yL = y**2
 
-    # if kind == "XSreduced":
-    #     return np.array([1.0, -yL / yp, f3sign * ym / yp])
-    # if kind == "XSyreduced":
-    #     return np.array([yp, -yL, f3sign * ym])
-
     # Neutral Currents
     # HERA average
     if kind == "XSHERANCAVG":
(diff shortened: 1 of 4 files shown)


src/yadism/esf/exs.py
```diff
@@ -5,7 +5,67 @@
...
+
+def xs_coeffs(kind, y, x=None, Q2=None, params=None):
+    """Compute coefficients in the definition of a given cross section.
+
+    Parameters
+    ----------
+    kind : str
+        the identifier of the cross section for which coefficients have to be
+        computed
+    y : float
```
felixhekhorn, line 20: Could you please provide some quick description? (since we're on it)

src/yadism/esf/exs.py
```diff
@@ -5,7 +5,67 @@
...
+    x: None or float
+    Q2: None or float
+    params : None or dict
+
+    """
+    yp = 1.0 + (1.0 - y) ** 2
+    ym = 1.0 - (1.0 - y) ** 2
+    yL = y**2
+
+    # if kind == "XSreduced":
```
felixhekhorn, line 30: Let's remove this code which we thought once upon a time we would maybe need - we accept the thing changes dataset by dataset

felixhekhorn (approved):
Apart from minor things this should be just a reshuffling and so ok ...

src/yadism/esf/exs.py
AleCandido: You're right, I'll do it (as you see I started, I simply forgot to complete).

src/yadism/esf/exs.py
AleCandido: Ok, I'll do.

Still thinking to improve the names, but for a *long* time being it is just fine.

felixhekhorn (approved):

Issue #122, opened 2021-10-26
Inherit hadronic target to PineAPPL


felixhekhorn:
in the PineAPPL generation we're hard coding at the moment the hadronic target: this might become a problem with NuTeV and CHORUS

https://github.com/N3PDF/yadism/blob/23c413de082d550b2161512b90a2157e2e69c3c2/src/yadism/output.py#L240

(the lepton is already correct:
https://github.com/N3PDF/yadism/blob/23c413de082d550b2161512b90a2157e2e69c3c2/src/yadism/output.py#L241
)

felixhekhorn:
> > Yet another question: for the previous CHORUS dataset you set `initial_state_1` to `2212`, so although this is supposed to be an antineutrino-lead scattering you require a proton PDF. I suppose that means that all nuclear effects are taken care of inside yadism, is my understanding correct?
> 
> yes, inside yadism we're applying "level-0" nuclear corrections, i.e.
> 
>     * adjust the electric charge
> 
>     * swap u and d
> 
> 
> this is related to [N3PDF/yadism#122](https://github.com/N3PDF/yadism/issues/122)
> 
> I don't think this is blocking at the moment ...
> 
> However, since we're speaking about it, I have a naive question: even with the correct initial state we still need to apply the "level-0" nuclear corrections inside `yadism`, right?

If you're taking care of 'translating the proton to the nuclear target' then everything's fine as far as I understand it.

_Originally posted by @cschwan in https://github.com/NNPDF/runcards/issues/117#issuecomment-1012168872_

felixhekhorn:
Let's move the discussion here, where it belongs ...

felixhekhorn:
> If you're taking care of 'translating the proton to the nuclear target' then everything's fine as far as I understand it.

well, that's the question:
- do we need to do 1 thing: (apply "level-0" nuclear correction + use initial_state=p) or (no nuclear-correction + initial_state = Fe)
- or do we need to do 2 things: apply "level-0" nuclear correction + initial_state = Fe

AleCandido:
This issue is getting old without nothing happening, nor going to happen soon.

@cschwan @felixhekhorn my personal point of view is:
- option 1 should be the default, not only because of PDF fitting (that is for sure relevant), but because it's easier you have proton PDFs and you're satisfied with level-0 corrections
- option 2 is always viable, without extra effort in `yadism`: you just need to pass a proton as a target in the runcard, and change later on the target inside the PineAPPL grid

[...]

PR #141, opened 2022-06-08
Move PineAPPL to yadbox

AleCandido:


felixhekhorn:
How can we retrigger the workflows?

AleCandido:
In theory you retrigger by going on the old workflow (in the actions tab), and clicking the button. But this commit was older than 6 months, so I believe the workflow have been canceled.

I had to rebase in any case, so now they have been re-triggered by the new commit :)

AleCandido:
It looks like we are not tracking `yadbox` (and probably nor `yadmark`) in the coverage...

felixhekhorn:
> It looks like we are not tracking `yadbox` (and probably nor `yadmark`) in the coverage...

looks like ... keep in mind to also adjust `pinecardsrunner` if you make a new release ...

AleCandido:
> looks like ... keep in mind to also adjust `pinecardsrunner` if you make a new release ...

Luckily Poetry is serving us well: I could make a new release of `yadism` without breaking `runcardsrunner` (even though this was actually safe), since the version is pinned.
But yes, the moment I'll do `poetry update` in `runcardsrunner` with this I'll have to update the code as well :)

felixhekhorn:
Feel free to merge when you're ready

AleCandido:
> Feel free to merge when you're ready

Almost: adding `yadbox` to the coverage doubled the codecov reports in the checks, but it seems like they refer always to `yadism` only, so I'd like to understand if it is actually working or not...

In any case, soon I will merge :)

AleCandido:
Actually, the double Codecov thing solved on its own, even though `yadbox` still seems to have no coverage.

At this point I would merge, and worry later about `yadbox` coverage (that is not top priority in any case).

felixhekhorn:
in the past we said we can merge, is this still the case? then we should do it @AleCandido

AleCandido:
We didn't much, I'm fairly confident it should work. But if you could test it (once and in a simple way, not a unit test) would be great!

felixhekhorn:
I think this is good to merge, do you agree? @AleCandido 

It will trigger 0.12 since we're breaking the API, i.e. `pinefarm` has to be adjusted

felixhekhorn:
by the way: the branch name is completely misleading - since this PR does not tackle #122

AleCandido:
> I think this is good to merge, do you agree? @AleCandido

Agreed

> It will trigger 0.12 since we're breaking the API, i.e. `pinefarm` has to be adjusted

The moment we want to release to `master`, but there is no urgent need.

> by the way: the branch name is completely misleading - since this PR does not tackle #122

I have the feeling you did it in the first place (since you authored the commit, I just rebased later...).
However, I guess the initial idea was to do it. Simply, this PR has been left behind for a while, and the initial goal has not been reached. But it is fine to merge, and open a new one (if someone wants to address #122).

PR #142, opened 2022-06-09
Fix anomalous dimension labels in eko 0.9.4

felixhekhorn:
note that this is not a problem in the released version as this is not happening in 0.9.1 (but indeed is happening in 0.9.4)


AleCandido:
Unfortunately, v0.9.4 is compatible with v0.9.1 (so it follows the rules of '^'). So, no problem for `yadism` environment, but for `yadism` dependencies, they will have to manually pin the `eko` version as well.

However, that's a good thing to remember for the future, it would have been a solution even without fixing this.

Before merging I'll try to fix the tests (otherwise we won't have a release in any case).

felixhekhorn:
I agree it was an oversight on the `eko` side (however, in some sense this is the price we're paying in not doing a gigantic QED PR, since we're constantly breaking (small) things ... meaning: e.g. https://github.com/N3PDF/eko/pull/115 should trigger a minor step)

felixhekhorn:
> Before merging I'll try to fix the tests (otherwise we won't have a release in any case).

ah sorry, yes - thanks! (same problem of course ...)

AleCandido:
Last run unfortunately is just bad: PyPI is currently down...

I will rerun tomorrow.

AleCandido:
We're back to a green tick, I'd propose to merge :)

AleCandido (approved):

PR #145, opened 2022-06-30
N3LO light coefficients

giacomomagni:
Implementation of light N3LO coefficients


AleCandido:
@giacomomagni I updated Pylint configurations from those in EKO.

In this case, since we are generating code (or at least converting), Pylint is invaluable to detect gross errors (like things undefined). Still, it's not a bad idea to manually review. I'll do it :)

src/yadism/coefficient_functions/light/f3_nc.py
```diff
@@ -17,13 +22,25 @@ def NLO():
 
     def NNLO(self):
         """
-        |ref| implements :eqref:`4.8`, :cite:`vogt-f2nc`.
+        |ref| implements the difference between :eqref:`2.8` and :eqref:`3.5, :cite:`Davies:2016ruz`.
```
giacomomagni, line 25: This should be the correct reference for this coefficient function, but I'm a bit confused why this contribution is not directly added in `f3_cc.NonSingletEven` only.

AleCandido:
> Is there any way we can test these numbers, beside comparing to APFEL/QCDNUM ?

N3LO is genuinely new, I don't believe you can compare with anyone.

The only thing we might have access to is the scale variation part, since should be proportional to lower orders, so if you manage to select that one only you can compare.

But the very interesting part is new, and I don't even know if there is any kind of sum rule we can check...
(otherwise they would be condition for structure functions, and we would have used for the neutrino project)

giacomomagni:
I added some benchmark against the new computation of the 3 loop coefficient function [here](https://arxiv.org/pdf/hep-ph/0504242.pdf). Overall the agreement is good apart from: 

* the local parts seem slightly to be different and this should be due to the fact that the previous results are approximated and contains some shifts in the local part to fine tune the accuracy as explicitly stated in the Fortran files. Instead the new one are exact.  
* Theres is a quite large contribution due to `d33` color factor that seems not included in the `FL` parametrisation. This still deserves some investigation

extras/light/n3lo_source/xcdiff3p.f
```diff
@@ -0,0 +1,133 @@
+*
```
felixhekhorn, line 1: Where is this file actually coming from? Looks almost like Vogt, but I can't see on his website ...

src/yadism/coefficient_functions/light/f3_cc.py
```diff
@@ -1,22 +1,51 @@
 # -*- coding: utf-8 -*-
+"""
+At |N3LO| that the source files in fortran follows :cite:`vogt-f3cc` notation
```
felixhekhorn, line 3:
```suggestion
Note that at |N3LO| the source files in fortran follow :cite:`vogt-f3cc` notation
```

src/yadism/coefficient_functions/light/f3_nc.py
```diff
@@ -1,7 +1,13 @@
 # -*- coding: utf-8 -*-
+"""
+At |N3LO| that the source files in fortran follows :cite:`vogt-f3cc` notation
```
felixhekhorn, line 3: I'd say let's not repeat the comment here, but let's refer just to the CC file.

felixhekhorn (approved):
- the actual expression are hard to check and the best we can do, is what you did - compare to a second source
- did I see correctly that e.g. src/yadism/coefficient_functions/light/n3lo/xc2ns3p.py is not fully auto-generated? which is fine, but maybe add a note somewhere so we can reconstruct the manual changes

giacomomagni:
> * did I see correctly that e.g. src/yadism/coefficient_functions/light/n3lo/xc2ns3p.py is not fully auto-generated? which is fine, but maybe add a note somewhere so we can reconstruct the manual changes

Yes this is correct, since I had to manually implement the opposite combination for charged currents, which was computed in the 2016 paper. I added a comment. 
I would like to solve the positivity bug (one point being empty) before merging this, or at least be sure it's not coming from the coefficients. can we merge #146 ?

Issue #151, opened 2022-09-06
Refactor `PTO` to `order` as in EKO?


felixhekhorn:
- @giacomomagni mirrored in #145 the theory card refactor from eko and introduced the `order` field
- I wonder whether instead we should call `eko.compatibility` inside and then act further (note that we need the eko settings for `Couplings.from_dict`
- as far as yadism is concerned the second element of `order` is never relevant (since QED corrections are out of scope, as discussed), instead the tuple `(PTODIS, PTO)` is relevant

felixhekhorn:
We need some compatibility between eko and yadism, since they share some settings and, more relevant, some code: `Couplings` (which explicitly depends on QED) - that's why I suggested to build on top of `eko.compatibility`

AleCandido:
> We need some compatibility between eko and yadism, since they share some settings and, more relevant, some code: `Couplings` (which explicitly depends on QED) - that's why I suggested to build on top of `eko.compatibility`

Not that sure: I'd like runcards to be independent, because there are a lot of theory parameters that are DIS only and Evolution only, and they should not enter each other cards (since we are not sensitive to it, they should not be provided).
Moreover, `yadism` is performing a different job than `eko`, the dependency is just an implementation detail (to save us some duplication), but should be kept internal: from the point of view of the user, they are separate and that's it.

Also consider that `compatibility` is a temporary layer, going to vanish as soon as possible (if we need compatibility with database, it should be dealt at the database interface level - i.e. in `pineko`).

giacomomagni:
I'm happy with this solution: 
```
# EKO
order: ["QCD", "QED"]
# yadism
order: "QCD"
evolution_order:`["QCD", "QED"]`
```
apart that `"QCD"` values will not have the same meaning, since we have a +1 for eko ...

AleCandido:
> apart that `"QCD"` values will not have the same meaning, since we have a +1 for eko ...

Yes, you're right wrt the pre-QED status. But after QED modification, they should be the exact same, i.e. the number of powers of $\frac{\alpha_s}{4\pi}$ (since we needed also the case with no QCD, and pure QED, i.e. `[0, 1]`).

PR #154, opened 2022-09-27
Update eko

felixhekhorn:
- I hope I caught everything
- @giacomomagni do you have by chance still xspace_bench in your env and can quickly run a benchmark? Otherwise, let's not worry for the moment, I'd say
- there is still the problem with `setuptools` (see also https://github.com/N3PDF/banana/issues/27): poetry removes it, but then the package fails (the error was again `No module named 'pkg_resources'`)


giacomomagni:
Hi Felix, I'll have check if I still have it installed...
I think we can merge #145, where some of the changes are already addressed
(my bad not having done it before...) is it okay ?

felixhekhorn:
With #145 in the main difference is that I used the eko compatibility layer always since `Couplings` is an eko object and we can't guarantee at the moment the yadism compatibility layer acts the same (see also #151)

PR #155, opened 2022-10-07
N3LO cc fix

giacomomagni:
Prompted @andreab1997 and his comparison of N3LO-NNLO shift with MHOU from scale variations, I checked the implementation of CC coefficients and find that there was a mistake in the usage of the color factors `fl11, fl02`. 

I sent and email to Joshua Davis the author of the main [reference](https://arxiv.org/pdf/1606.08907.pdf) and this was his reply: 

```
Hi Giacomo,
As far as I remember the notation, the fl11 diagrams should never be
included in any CC DIS, whether you consider nu+nubar or nu-nubar.

This class has the two scattering bosons coupling to different fermion
lines (Fig 1 of https://arxiv.org/pdf/hep-ph/0504242.pdf ) and so when
the bosons are charged these diagrams are forbidden since only gluons
connect them.

In addition, fl02 diagrams don't contribute to C3 for nu-nubar DIS, and
so those are removed from the nu+nubar C3 when forming delta C3.
Let me know, if this doesn't make sense!

Thanks, Josh.
```

I believe this is now fixed accordingly, but still the comparison produced by Andrea looks odd (the label should be N3LO-NNLO shift) ...

image
 
 So there is still some investigation do to...

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -6,12 +6,6 @@ on:
       - "*docs*"
 
 jobs:
-  py37:
-    uses: N3PDF/workflows/.github/workflows/python-poetry-tests.yml@main
-    with:
-      python-version: "3.7"
-      poetry-extras: "-E mark"
-
   py38:
     uses: N3PDF/workflows/.github/workflows/python-poetry-tests.yml@main
     with:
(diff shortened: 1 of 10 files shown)


felixhekhorn (approved):
I agree the plot still looks a bit weird, but I don't have a clear suggestion on what to improve ... so maybe fine and we can re-asses at a later point

PR #157, opened 2022-11-02
Make yadism compatible with eko 0.11

andreab1997:
At the end, for the `convolute_operator` function, it seems that changing `xgrid_raw` to `xgrid.raw` was enough. Anyway I have opened this PR in case I will find other things to be changed in the next future. @AleCandido @felixhekhorn


AleCandido (approved):
The change is minimal and needed. I didn't run myself, but I agree that this change has to be done

AleCandido (changes requested):
Sorry if I change my mind (slightly).
Actually, it is currently broken, since you kept:
https://github.com/NNPDF/yadism/blob/a7dea9a2c072dd82780b1238e7a59f218dda85a1/pyproject.toml#L43

Update it accordingly, and run `poetry update`. Then you can merge :)

tests/yadism/test_tmc.py
```diff
@@ -159,10 +159,10 @@ def isdelta(pdf):  # assert F2 = pdf
             isdelta(pdf_lin)
             # int_const = int_xi^1 du = 1-xi
             integral_with_pdf = np.matmul(res_const.orders[lo][0][0], pdf_lin)
-            assert pytest.approx(integral_with_pdf, 1 / 1000.0) == c * (1.0 - obj.xi)
+            np.testing.assert_allclose(integral_with_pdf, c * (1.0 - obj.xi), rtol=3e-2)
```
giacomomagni, line 162: I have no clue why I had to change this in order to pass...

AleCandido: If you want, you can add `rtol` also using `pytest.approx`. We are using `np.testing`  mainly when arrays are involved, but [`pytest.approx`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-approx) actually works also in that case (so it is a complete replacement)

However, yes, it is weird...

tests/yadism/test_output.py
```diff
@@ -25,6 +25,7 @@ def xfxQ2(self, pid, x, Q2):
 class TestOutput:
     def fake_output(self):
         out = dict()
+        out["xgrid"] = {"grid": np.array([0.5, 1.0]).tolist()}
         out["interpolation_xgrid"] = np.array([0.5, 1.0])
```
AleCandido, line 29: Now there are two `xgrid`s (`out["xgrid]` and `out["interpolation_xgrid"]`) keep a single one

tests/yadism/test_tmc.py
```diff
@@ -159,10 +159,10 @@ def isdelta(pdf):  # assert F2 = pdf
...
             # int_h2 = int_xi^1 du/u = -ln(xi)
             integral_with_pdf = np.matmul(res_h2.orders[lo][0][0], pdf_lin)
-            assert pytest.approx(integral_with_pdf, 1 / 1000.0) == c * (-np.log(obj.xi))
+            np.testing.assert_allclose(integral_with_pdf, c * (-np.log(obj.xi)), rtol=3e-2)
```
AleCandido, line 165: And this is weird as well (but essentially the same of above...)

AleCandido (changes requested):

AleCandido:
At this point I'd even approve the PR, but if possible it would be interesting to know more about the precision in the TMC integration.

In principle, it has not been touched (just two characters), so nothing should changed...
And an increment in tolerance can mask an actual bug. Since only the xgrid has been changed, I'd point to something like wrong usage of xgrid and its logarithm (something like `xgrid.raw` vs `xgrid.grid`), but in principle almost nothing has changed, so it is still weird...

giacomomagni:
> In principle, it has not been touched (just two characters), so nothing should changed... And an increment in tolerance can mask an actual bug. Since only the xgrid has been changed, I'd point to something like wrong usage of xgrid and its logarithm (something like `xgrid.raw` vs `xgrid.grid`), but in principle almost nothing has changed, so it is still weird...

I fear I'm not able to understand the origin of this... @andreab1997 might have a clue?
 I tried a benchmark with apfel:
```
tmc = BenchmarkTMC()
tmc.benchmark_nlo()
```

and indeed it's not working.

AleCandido:
If even benchmarks are broken, I suspect we are introducing a bug, so I'd block merging for as long as we do not understand what is going on. @felixhekhorn

giacomomagni:
> If even benchmarks are broken, I suspect we are introducing a bug, so I'd block merging for as long as we do not understand what is going on. @felixhekhorn

Sorry, I noticed only now that there was a comment about FL with TMC being broken in apfel, so I'd amend my statement, so `F2_total` and `F3_total` with `FONLL-A` and `TMC` don't display any major difference.

tests/yadism/test_tmc.py
felixhekhorn: Might it be that we change from linear interpolation to log interpolation? because now https://github.com/NNPDF/eko/blob/e3466fae35fd245c7e1a99158d8582f590276555/src/eko/interpolation.py#L501-L505 by default we're  log ... before there was the explict `log=False` argument (see https://github.com/NNPDF/eko/blob/01efee1d27d215e895dd055debff8905af3d548c/src/eko/interpolation.py#L439) in line 50

Note that linear here is fine, since we're faking in any case with 3 points in the linear regime. So to recover make `xg` linear, I guess ..

tests/yadism/test_tmc.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, tmc, xg=None):
         self.runner.configs = MockObj()
         self.runner.configs.M2target = 1.0
         self.runner.configs.TMC = tmc
-        self.runner.configs.interpolator = InterpolatorDispatcher(xg, 1, False, False)
+        self.runner.configs.interpolator = InterpolatorDispatcher(xg, 1, False)
```
felixhekhorn, line 50: this is the `log=False` I meant [here](https://github.com/NNPDF/yadism/pull/157/files#r1013810070)

tests/yadism/test_tmc.py
AleCandido: You're right. Then, I'd try to revert and check if the problem persists.
Can you do this? @andreab1997 @giacomomagni

src/yadism/runner.py
```diff
@@ -95,7 +95,10 @@ def __init__(self, theory: dict, observables: dict):
         self._observables = new_observables
 
         # Setup eko stuffs
-        interpolator = InterpolatorDispatcher.from_dict(self._observables, mode_N=False)
+        xgrid = XGrid(self._observables["interpolation_xgrid"])
```
felixhekhorn, line 98: Actually, even here: you should read the "logginess" from the run card ...

felixhekhorn (changes requested):

src/yadism/runner.py
AleCandido: Standard key has been `"interpolation_is_log"` until now. We need a rework of the runcard also for yadism, but not now

tests/yadism/test_tmc.py
giacomomagni: Thanks @felixhekhorn for the suggestion, indeed this solved the problem!!  :)

src/yadism/runner.py
giacomomagni: I should have solved it!

tests/yadism/esf/test_conv.py
```diff
@@ -119,7 +119,7 @@ def test_symmetric_conv(self):
     @pytest.mark.eko
     def test_basis_function_void(self):
         xg = np.linspace(0.2, 1.0, 5)  # 0.2, 0.4, 0.6, 0.8, 1.0
-        i = InterpolatorDispatcher(xg, 1, False, False)
+        i = InterpolatorDispatcher(xg, 1, False)
```
felixhekhorn, line 122: Apparently it doesn't matter, but still let's keep the old behaviour: so please also change here to linear

tests/yadism/test_structure_function.py
```diff
@@ -19,7 +19,7 @@ class MockObj:
 
 
 xg = np.linspace(0.2, 1.0, 5)  # 0.2, 0.4, 0.6, 0.8, 1.0
-interpolator = InterpolatorDispatcher(xg, 1, False, False)
+interpolator = InterpolatorDispatcher(xg, 1, False)
```
felixhekhorn, line 22: also please here back to linear

felixhekhorn (approved):

PR #161, opened 2022-12-02
Add an observable relevant for FPF studies

Radonirinaunimi:
For the FPF studies we need a definition of the double differential cross section that is not available yet. This PR just adds such a definition.

### DIFF
--- a/src/yadism/esf/exs.py
+++ b/src/yadism/esf/exs.py
@@ -5,7 +5,6 @@ from ..observable_name import ObservableName
 from .esf import ESFInfo
 from .result import ESFResult, EXSResult
 
-INV_GEV_TO_PB = 3.8937966e8 # pb
 GEV_CM2_CONV = 3.893793e10
 "Conversion factor from GeV^-2 to 10^-38 cm^2"
 
@@ -69,6 +68,7 @@ def xs_coeffs(kind, y, x=None, Q2=None, params=None):
         yp = 1.0
         yL = y**2 / (2 * (y**2 / 2 + (1 - y) - (mn * x * y) ** 2 / Q2))
     elif kind == "XSFPFCC":
+        INV_GEV_TO_PB = GEV_CM2_CONV / 100.0 # Pb
         norm = (INV_GEV_TO_PB * params["GF"] ** 2) / (4.0 * np.pi)
         norm *= 1.0 / (2.0 * x * (1.0 + Q2 / params["M2W"])**2)
     else:
--- a/src/yadism/esf/exs.py
+++ b/src/yadism/esf/exs.py
@@ -5,6 +5,7 @@ from ..observable_name import ObservableName
 from .esf import ESFInfo
 from .result import ESFResult, EXSResult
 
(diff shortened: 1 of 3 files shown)


src/yadism/esf/exs.py
```diff
@@ -5,6 +5,7 @@
 from .esf import ESFInfo
 from .result import ESFResult, EXSResult
 
+INV_GEV_TO_PB = 3.8937966e8 # pb
```
AleCandido, line 8: @Tanjona, the conversion between `pb` and `cm2` is almost trivial, and in any case I would store in a separate constant.

As you can see, we have now two constants, with a mantissa that is the same up to $10^{-6}$, and the difference is just the exponent (so you multiply this by `1e2` and you obtain essentially next line...)

AleCandido:
Benchmarks are failing because they should not run, I lost track where they have been introduced (there is a long discussion going on since 2 years with @felixhekhorn, I hope I will fix this, sooner or later...).

Apart the comment above, the rest looks good to me.

When you are ready, just ask for review :)

src/yadism/esf/exs.py
Radonirinaunimi: This does make sense! I have removed the explicit definition of the constant but rather kept the variable name in the code just to make it explicit which conversion is being done XD.

Radonirinaunimi:
> Benchmarks are failing because they should not run, I lost track where they have been introduced (there is a long discussion going on since 2 years with @felixhekhorn, I hope I will fix this, sooner or later...).
> 
> Apart the comment above, the rest looks good to me.
> 
> When you are ready (i.e. replied to the comment), just ask again for review and I will approve :)

I was indeed wondering about the failing benchmarks but knew that this surely had nothing to do with the modifications.

AleCandido (approved):

Radonirinaunimi:
Thanks! I will merge this then now.

src/yadism/esf/exs.py
```diff
@@ -67,6 +67,10 @@ def xs_coeffs(kind, y, x=None, Q2=None, params=None):
         ym = 0
         yp = 1.0
         yL = y**2 / (2 * (y**2 / 2 + (1 - y) - (mn * x * y) ** 2 / Q2))
+    elif kind == "XSFPFCC":
+        INV_GEV_TO_PB = GEV_CM2_CONV / 100.0 # Pb
```
felixhekhorn, line 71: I guess the comment should be rather `pb` (note the small p - as in pico) (just to not confuse with lead :upside_down_face: )

AleCandido: To really nitpick `Pb` in this context would mean "PetaBarn", instead of "picoBarn", according to SI.

Radonirinaunimi: This is definitely a fair point! In addition, in what we are doing we also need the observable name added to the `observable_names.yaml`. Let me restore this branch and add these changes here.

Radonirinaunimi: Now in #163.

PR #168, opened 2023-01-25
[DEPRECATED] FONLL matching away from threshold

felixhekhorn:
While checking the charm contributions we started to delay the bottom patch, i.e. use matching away from mass threshold. This revealed a bug in the implementation. In fact this feature never got correctly implemented (see also https://app.clickup.com/t/j3895p) .

- we did correctly account for the different evolution length by constructing our threshold atlas with the shifted ratios: https://github.com/NNPDF/yadism/blob/5c52c1a953def26fc8a9ec549ad0c2c8978d23d2/src/yadism/runner.py#L120
- but we did not correctly account for the shift in the matching conditions, e.g. here: https://github.com/NNPDF/yadism/blob/5c52c1a953def26fc8a9ec549ad0c2c8978d23d2/src/yadism/coefficient_functions/fonll/partonic_channel.py#L81
- As said there in the docstring it implements Eq. (100) of the [NPB](https://doi.org/10.1016/j.nuclphysb.2010.03.014) version - however as said below Eq. (30) of that paper, it should implement Eq. (B4) of [this paper (another Buza et al)](https://arxiv.org/pdf/hep-ph/9612398.pdf) . To be precise: the terms proportional $\ln(\mu^2/m^2)$ are missing and we need them. They account for the evaluation of the (true) matching conditions away from threshold that is compensating for the different evolution legth
- for the intrinsic case we were almost doing it, since here we followed Eq. (20a+20b) of the  [PLB](https://doi.org/10.1016/j.physletb.2015.12.077) version - except that the variable called $Q^2$ there is _not_ the virtuality of the photon, but is the matching scale $\mu^2$. This bug is fixed in https://github.com/NNPDF/yadism/commit/cf1a828ceb1a48bf83641419e51c6a8b89e4b33b
- ~~while looking at this I realized that $K_{gh}$ (Eq. (20b)) is not implemented correctly - this is fixed in https://github.com/NNPDF/yadism/commit/3c4a95ab8b3e7d220ce063432c5dd513fade3553~~ aaa, this is not Eq. (20b), most likely is correct as is (commit is reverted); however I'm almost sure we need to do adjust the `l` definition which also effects this part
- the eko counterparts are [here](https://github.com/NNPDF/eko/blob/e3466fae35fd245c7e1a99158d8582f590276555/src/eko/matching_conditions/as1.py) and there are L around (and they are the ratio $\mu^2/m^2$)

I still need to do:
- [ ] Fix [K_qq](https://github.com/NNPDF/yadism/blob/5c52c1a953def26fc8a9ec549ad0c2c8978d23d2/src/yadism/coefficient_functions/fonll/partonic_channel.py#L81)


felixhekhorn:
Close in favor of #170

PR #170, opened 2023-02-03
FONLL matching away from mass

felixhekhorn:
#168 provides a too complicated and potentially wrong solution to the problem at hand

It should be sufficient to just use the correct log, which is $Q^2/m^2$ and no $\mu$ - this instead can only appear in the atlas.

### TODO
- [x] drop mention of mu2 inside coefficient functions
- [x] write some documentation


AleCandido:
@felixhekhorn why this is still draft?

felixhekhorn:
I suggest to move

>  drop mention of mu2 inside coefficient functions

in a separate PR

AleCandido:
> I suggest to move
> 
> > drop mention of mu2 inside coefficient functions
> 
> in a separate PR

Sorry, I already did it, but on the wrong branch (too similar names).

AleCandido (approved):
The only item I'm missing, that I'd like to still review, is the docs, but this I will read with ease from RTD after merged. And I'd not block the PR on that.

PR #171, opened 2023-02-10
Upgrade to EKO v0.12

AleCandido:
Things to be fixed

- [x] `.from_dict()` dropped
- [x] `compatibility` module dropped


pyproject.toml
```diff
@@ -100,8 +100,8 @@ yadnav = "yadmark.navigator:launch_navigator"
 coverage = "$BROWSER htmlcov/index.html"
 test = "pytest tests"
 test-cov = ["test", "coverage"]
-lint = "pylint src/*/*.py -E"
-lint-warnings = "pylint src/*/*.py --exit-zero"
+lint = "pylint src/**/*.py -E"
+lint-warnings = "pylint src/**/*.py --exit-zero"
```
AleCandido, line 104: Actually, I would propose to drop the path from here all at once, and just use the configuration files.

Take into account that for `bash` they are the same, unless you enable `shopt -s globstar`. So it is shell and configuration dependent, while I believe that `src/` is just enough (and you can put in configs)

AleCandido: Or, better, you can ignore paths:
https://pylint.readthedocs.io/en/latest/user_guide/configuration/all-options.html#ignore-paths

But at a first run I would leave it empty, just drop the path here:
```suggestion
lint = "pylint -E"
lint-warnings = "pylint --exit-zero"
```

felixhekhorn: if you drop all path nothing happens ...

AleCandido: I confused with Pytest

felixhekhorn:
@AleCandido if you agree we can merge+tag+release and tick @andreab1997 to restart

AleCandido:
I'm the PR author, so I don't have the approve button, but... I approve!

felixhekhorn:
Question: why don't we have a
```
    with:
      poetry-extras: "-E mark -E pineappl"
```
inside the pypi workflow? but we have something similar in eko ...

PR #173, opened 2023-02-22
Implement the light part of the polarized coefficient functions

Radonirinaunimi:
Implement the massless spin-dependent coefficient functions $g_1$, $g_4$ and $g_L$ (up to NNLO).

Status of the **NC** implementation:
| SF | LO | NLO [NS, Gluon]  | NNLO [NS, Gluon, S]  |
|----|----|-----|-----|
| g1 | ✅ | [✅, ✅]  | [✅, ✅ , ✅] |
| g4 | ✅ | [✅, 0]   | [✅, 0 , 0] |
| gL |  0  | [✅, 0]   | [✅, 0 , 0] |

TODO: 
* [x] Apfel++ benchmark and parity violating couplings fixes.
* [x] Target Mass Correction (implemented for $g_1$)
* [ ] ~CC implementation (should be straight forward) same relations ad NC~ (Omitted)
* [ ] ~TMC for polarized observables~ (delegated to #187)

### DIFF
--- a/.github/workflows/regression.yml
+++ b/.github/workflows/regression.yml
@@ -1,9 +1,10 @@
 name: regression
 
 on:
-  push:
-    branches-ignore:
-      - "*docs*"
+  workflow_dispatch:
+  # push:
+  #   branches-ignore:
+  #     - "*docs*"
 
 jobs:
   regression:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
       - id: isort
         args: ["--profile", "black"]
   - repo: https://github.com/asottile/pyupgrade
-    rev: v3.3.1
+    rev: v3.4.0
     hooks:
       - id: pyupgrade
   - repo: https://github.com/hadialqattan/pycln
@@ -44,6 +44,6 @@ repos:
         additional_dependencies:
           - toml
   - repo: https://github.com/pre-commit/pre-commit
-    rev: v3.2.2
+    rev: v3.3.1
     hooks:
       - id: validate_manifest


src/yadism/observable_name.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 
 fake_kind = "??"
-sfs = ["F2", "FL", "F3"]
+sfs = ["F2", "FL", "F3", "G1"]
```
felixhekhorn, line 4:
```suggestion
sfs = ["F2", "FL", "F3", "g1"]
```
maybe?

giacomomagni:
Benckmarks with Apfel:

| SF | LO | NLO |
|----|----|-----|
| g1 | ✅   |  ✅   |   |   |
| g4 |    |     |   |   |
| gL |    |     |   |   |

src/yadism/coefficient_functions/light/nlo/g1.py
```diff
@@ -0,0 +1,27 @@
+import numba as nb
+import numpy as np
+from eko.constants import TR, CF
+
+from . import f2
```
Radonirinaunimi, line 5:
```suggestion

```

src/yadism/coefficient_functions/light/nlo/g1.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@nb.njit("f8(f8,f8[:])", cache=True)
+def ns_reg(x, args):
+    return (
+        2
+        * CF
+        * (-(1 + x) * np.log(1 - x) - (1 + x**2) * np.log(x) / (1 - x) + 2.0 + x)
+    )
```
Radonirinaunimi, line 14:
```suggestion
def ns_reg(z, args):
    return (
        2
        * CF
        * (-(1 + z) * np.log(1 - z) - (1 + z**2) * np.log(z) / (1 - z) + 2.0 + z)
    )
```
Maybe for consistency let's use $z$ everywhere.

src/yadism/coefficient_functions/light/nlo/g1.py
giacomomagni: it's still used for the local and singular part.

Radonirinaunimi: My bad! Missed it by looking at the commit only.

src/yadmark/benchmark/external/apfel_utils.py
```diff
@@ -29,6 +29,7 @@ def load_apfel(theory, observables, pdf="ToyLH", use_external_grid=False):
     apfel.SetProcessDIS(observables.get("prDIS", "EM"))
     apfel.SetPropagatorCorrection(observables.get("PropagatorCorrection", 0))
     apfel.SetPolarizationDIS(observables.get("PolarizationDIS", 0))
+    apfel.SetPolarizedEvolution(observables.get("PolarizationDIS", 0))
```
felixhekhorn, line 32: Attention! this is not a boolean, but a float - setting the amount of leptonic polarization (and only that) https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/src/yadism/coefficient_functions/coupling_constants.py#L107
instead `SetPolarizedEvolution` refers to the hardon/proton side

src/yadism/coefficient_functions/light/nlo/g1.py
```diff
@@ -1,18 +1,20 @@
...
 import numpy as np
-from eko.constants import TR
-
+from eko.constants import TR, CF
 
 from . import f2
 
 
 @nb.njit("f8(f8,f8[:])", cache=True)
 def ns_reg(z, args):
```
felixhekhorn, line 9:
```suggestion
def ns_reg(z, _args):
```

src/yadism/coefficient_functions/light/g1_nc.py
```diff
@@ -0,0 +1,52 @@
...
+from ..partonic_channel import RSL
+from . import n3lo, nlo, nnlo
+from . import partonic_channel as pc
+
+
+class NonSinglet(pc.LightBase):
+    @staticmethod
+    def LO():
+        """
+        |ref| APFEL
```
felixhekhorn, line 11: please find a proper reference

src/yadmark/benchmark/external/apfel_utils.py
giacomomagni: uhh ops, I was not aware of this. So the fix would be to set `SetPolarizedEvolution ` according to the observable name? Or would you prefer to have another key in the ocard ?

AleCandido: I agree, I'd make it dependent on the observable name

giacomomagni: okay then we go for the first option

src/yadism/observable_name.py
```diff
@@ -1,7 +1,5 @@
-# -*- coding: utf-8 -*-
-
 fake_kind = "??"
-sfs = ["F2", "FL", "F3", "g1"]
+sfs = ["F2", "FL", "F3", "g1", "gl"]
```
felixhekhorn, line 2: I would suggest `gL` (as e.g. in PDG)

giacomomagni: I agree, also more consistent with `FL`: 
```suggestion
sfs = ["F2", "FL", "F3", "g1", "gL"]
```

src/yadmark/benchmark/external/apfel_utils.py
```diff
@@ -25,10 +24,18 @@ def load_apfel(theory, observables, pdf="ToyLH", use_external_grid=False):
         theory, observables, pdf, use_external_grid=use_external_grid
     )
 
+    is_polarized = ["g" in obs_name for obs_name in observables["observables"]]
```
giacomomagni, line 27: need to be fixed, see [here](https://github.com/NNPDF/yadism/pull/180#discussion_r1135285051)

src/yadism/coefficient_functions/light/nlo/g1.py
```diff
@@ -0,0 +1,27 @@
+import numba as nb
+import numpy as np
+from eko.constants import TR, CF
+
+from . import f2
+
+
+@nb.njit("f8(f8,f8[:])", cache=True)
+def ns_reg(z, _args):
```
giacomomagni, line 9: @Radonirinaunimi here we should import this piece from `f3.ns_reg` eq 19 and paragraph below of [here](https://arxiv.org/pdf/2210.12014.pdf)

Radonirinaunimi: I could not find from the cited references the explicit equation for such an equivalence relation, but upon doing such a change the results are exactly the same as before so this is good enough for me.

Since these equivalence relations are valid up to NNLO it would be even much easier to implement the NNLO $g_1$ contributions as we will not have to implement the lengthy equations in the old paper.

src/yadmark/benchmark/external/apfel_utils.py
AleCandido: `obs_name.startswith("g")`

Radonirinaunimi: Yes, this will be fixed by @RoyStegeman in #180 after which we'll merge the changes here.

src/yadism/coefficient_functions/light/nnlo/g1.py
```diff
@@ -53,41 +65,82 @@ def gluon_reg(z, args):
...
     ln1pz = np.log(1 + z)
     ln1pz2 = ln1pz**2
-    # fmt: off
-    return nf * (
-        CF * TR * (
-            ( 1 - 2 * z ) * ( 
-                32 * Li31mz - 16 * ln1mz * Li21mz - 8 * lnz * Li21mz - 24 * zeta2 * lnz
-                - 20 * ln1mz3 / 3.0+ 16 * lnz * ln1mz2 - 16 * lnz2 * ln1mz + 10 * lnz3 / 3
+
+    gluon_regular = nf * (
```
AleCandido, line 69: This variable is not used, other than returning it, and it is the exact name of the function.

Since this is not Fortran code, feel free to immediately return:
```suggestion
    return nf * (
```

Radonirinaunimi: Since the expressions are so long, having a variable would be more helpful in terms of debugging, but fine given that it is just a copy & paste I guess that won't hurt.

src/yadism/coefficient_functions/light/nnlo/g1.py
```diff
@@ -53,41 +65,82 @@ def gluon_reg(z, args):
...
-        + CF * TR * ( 
-            16 * ( 1 + 2 * z ) * ( 2 * Li21mz + 2 * lnz * ln1mz - lnz2 )
-            + 96 * ( 1 - z ) * ln1mz - ( 144 + 64 * z ) * lnz - 304 * ( 1 - z )
+        + CF
+        * TR
+        * (
+            16 * (1 + 2 * z) * (2 * Li21mz + 2 * lnz * ln1mz - lnz2)
+            + 96 * (1 - z) * ln1mz
+            - (144 + 64 * z) * lnz
+            - 304 * (1 - z)
```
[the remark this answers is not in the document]
Radonirinaunimi: Done! This was indeed forgotten.

giacomomagni:
> The implementation up to NNLO should now be complete. Now, we should test and/or benchmark these.

Okay I'll take care of extending `yadmark` to run `apfelxx`. 
The installation will be provided by [this PR](https://github.com/NNPDF/workflows/pull/12)

Radonirinaunimi:
> > The implementation up to NNLO should now be complete. Now, we should test and/or benchmark these.
> 
> Okay I'll take care of extending `yadmark` to run `apfelxx`. The installation will be provided by [this PR](https://github.com/NNPDF/workflows/pull/12)

Awesome! Thanks.

docs/source/refs/cf-light.bib
```diff
@@ -102,3 +101,65 @@ @article{Davies:2016ruz
...
+    title = "{APFEL++: A new PDF evolution library in C++}",
+    eprint = "1708.00911",
+    archivePrefix = "arXiv",
+    primaryClass = "hep-ph",
+    doi = "10.22323/1.297.0201",
+    journal = "PoS",
+    volume = "DIS2017",
+    pages = "201",
+    year = "2018"
+}
```
RoyStegeman, line 165:
```suggestion
}

```

src/yadism/coefficient_functions/light/gl_nc.py
```diff
@@ -0,0 +1,31 @@
...
+        return RSL(
+            nnlo.xclns2p.clnn2a, loc=nnlo.xclns2p.clnn2c, args=dict(reg=[self.nf])
+        )
+
+
+class Gluon(epc.EmptyPartonicChannel):
+    pass
+
+class Singlet(epc.EmptyPartonicChannel):
+    pass
```
RoyStegeman, line 31:
```suggestion
    pass

```

Radonirinaunimi:
> Do we need to extend TMC also for polarised observables??

Yes, I think so, given that the expressions are available. But that we might want to do in a separate PR and probably after #181 is merged.

src/yadism/esf/tmc.py
```diff
@@ -479,7 +537,66 @@ def _get_result_exact(self):
...
+        # Call to the raw integrals
+        k1out = self._k1()
+        k2out = self._k2()
+        # Combine the expressions
+        return self._factor_shifted * g1out + self._factor_k1_k2 * (
+            self._factor_k1 * k1out + self._factor_k2 * k2out
+        )
+
+    def _get_result_APFEL(self):
+        """APFEL does not implement polarized TMC"""
```
giacomomagni, line 596:
```suggestion
        raise NotImplementedError("APFEL does not implement polarized TMC")
        
```

src/yadism/esf/tmc.py
```diff
@@ -46,22 +49,33 @@ def g2_ker(z, _args):
     return 1 - z
 
 
+@nb.njit("f8(f8,f8[:])", cache=True)
+def k1_ker(_z, _args):
+    # TODO: Check that following makes sense
+    return 1
```
Radonirinaunimi, line 55: @felixhekhorn This is what I was alluding to, although looking at the ration doesn't hint that something is wrong.

pred_g1_total

Radonirinaunimi:
> `extras/` is mostly preserving the historic memory of yadism development, to reconstruct how and why we did things. For sure at some point we might want to clean up things, but for the time being we are busy enough with higher priorities, so personally I will not invest much energy in it soon.

I understand, it was more of since I will be touching that folder I might as well do so. Just wanted to know if you'd agree that this should be taken care of in the future.

AleCandido:
> I understand, it was more of since I will be touching that folder I might as well do so. Just wanted to know if you'd agree that this should be taken care of in the future.

In the future for sure (something we'll want to keep, but definitely there are stuffs to be dropped).

For the time being, my advice is to just open a new folder inside, name after your topic, and put your stuffs in. Just forget about the rest :sweat_smile:

src/yadism/esf/tmc.py
```diff
@@ -282,6 +296,50 @@ def _g2(self):
         # as we get a 1/z by the measure and an evaluation of 1-xi/z
         return self._convolute_FX("F2", g2_ker)
 
+    def _k1(self):
```
felixhekhorn, line 299: since this is g1 specific you can also implement down there rather than in this base class

src/yadism/esf/tmc.py
```diff
@@ -282,6 +296,50 @@ def _g2(self):
...
+    def _k2(self):
+        r"""
+            Compute the raw integral that enters the computation of `g`
+            making use of :py:meth:`_convolute_FX`.
+
+            .. math::
+                :nowrap:
+
+                \begin{align*}
+                k_2(\xi,Q^2) &= \int_\xi^1 du \log(\frac{v}{\xi})
```
felixhekhorn, line 328: is this `v` a `u`?

src/yadism/esf/tmc.py
```diff
@@ -479,7 +537,67 @@ def _get_result_exact(self):
...
 
 
-ESFTMCmap = {"F2": ESFTMC_F2, "FL": ESFTMC_FL, "F3": ESFTMC_F3}
+class ESFTMC_g1(EvaluatedStructureFunctionTMC):
+    """
+    This function implements the actual formula for target mass corrections
+    for parity conserving polarized structure function g1, for the two kinds
+    described in the parent class :py:class:`EvaluatedStructureFunctionTMC`.
+
+    The formula in question can be found in :cite:`Accardi:2008pc`,
```
felixhekhorn, line 546: Can you please quote the exact equation? This makes my review easier :upside_down_face:  I guess it is Eq. 10 of `:cite:tmc-g1-iranian`, yes?

src/yadism/esf/tmc.py
```diff
@@ -479,7 +537,67 @@ def _get_result_exact(self):
...
+    def __init__(self, SF, kinematics):
+        super().__init__(SF, kinematics)
+        # Kinematic factor for ZM g1
+        self._factor_shifted = self.x**2 / (self.xi * self.rho**3)
+        # Kinematic factor for common for `k1` & `k2` integral
+        self._factor_k1_k2 = (4 * self.x**2 * self.mu) / self.rho**3
+        # Kinematic factor specific for `k1` integral
+        self._factor_k1 = (self.x + self.xi) / self.xi
+        # Kinematic factor specific for `k2` integral
+        self._factor_k2 = (self.rho**2 - 3) / (2 * self.rho)
```
felixhekhorn, line 568: Looking at Eq. 10 of `:cite:tmc-g1-iranian` I can somehow identify `_factor_k1`, but the 3 inside `_factor_k2` I can not identify ...

src/yadism/esf/tmc.py
```diff
@@ -479,7 +537,67 @@ def _get_result_exact(self):
...
+        # Call to the raw integrals
+        k1out = self._k1()
+        k2out = self._k2()
+        # Combine the expressions
+        return self._factor_shifted * g1out + self._factor_k1_k2 * (
+            self._factor_k1 * k1out + self._factor_k2 * k2out
+        )
+
+    def _get_result_APFEL(self):
+        raise NotImplementedError("APFEL does not implement polarized TMC")
```
felixhekhorn, line 596: Fair enough, the name is very bad - but this actually has nothing to do with APFEL, we're just referring to the APFEL approximation, i.e. only including one integral (and not the nested)

felixhekhorn (changes requested):
Convert comments on https://github.com/NNPDF/yadism/commit/9943e65694e4885f4b6c0c0c764e6c6c1f1a7d37#comments to review (sorry)

src/yadism/esf/tmc.py
felixhekhorn: Actually, there seems to be a bug for F3 (cc @AleCandido ):
https://github.com/NNPDF/yadism/blob/d947168752d48636a1c0fa0cc6709c27150b81bd/src/yadism/esf/tmc.py#L526-L528
but we're not doing what we promise:
https://github.com/NNPDF/yadism/blob/d947168752d48636a1c0fa0cc6709c27150b81bd/src/yadism/esf/tmc.py#L44
instead it should be `h3_ker = k1_ker`

In any case `_convolute_FX` should do a convolution as I said ...

src/yadism/esf/tmc.py
felixhekhorn: To be more specific looking at Eq. 20 of [0709.1775](https://arxiv.org/pdf/0709.1775.pdf) $g_2$ should also emerge from a double integral same as $k_2$ (although I can not find it straight), so here you should just drop $k_2$. (Physics wise they should refer to higher twist corrections ...)

src/yadism/esf/tmc.py
Radonirinaunimi: Perhaps it might be easier to refer to Eq. 26 of `tmc-g1-accardi` as the expression there is more compact. From there, the $3$ stems from replacing $\gamma^2 -  2 \longrightarrow ((1 + \gamma^2 \to \rho^2) - 3)$.

src/yadism/esf/tmc.py
Radonirinaunimi: Hadn't noticed that you were already suggesting this! This is done done.

Radonirinaunimi:
Are you happy with these now @felixhekhorn?

src/yadism/esf/tmc.py
felixhekhorn: @Radonirinaunimi can you double check this? and fix if you agree

felixhekhorn (approved):

felixhekhorn:
apart from the potential $F_3$ bug this should be fine ...

giacomomagni:
> apart from the potential F3 bug this should be fine ...

Do you mean `g1 == F3` for the NS ?

Radonirinaunimi:
> > apart from the potential F3 bug this should be fine ...
> 
> Do you mean `g1 == F3` for the NS ?

No, I believe Felix meant this comment:

https://github.com/NNPDF/yadism/pull/173/files#r1183519543

AleCandido:
Part of the breaking tests are breaking because we do not support py3.7 any longer, at least in EKO (and consequently also in yadism, at least because of the dependency).

It should be sufficient to upgrade the version range of *regression* to be the same of *unit tests*.

However, even py3.8-9 are failing, and here the workflow complains about a missing `requirements.txt` file.
Should we disregard these tests at all (then deactivating), solve here, or solve in a further PR (i.e. open now an issue)?

benchmarks/runners/apfel_.py
```diff
@@ -280,9 +302,9 @@ def benchmark_nlo(self):
     # proj.benchmark_pto(0)
     # proj.benchmark_pto(1)
 
-    pos = BenchmarkPositivity()
-    pos.benchmark_pto(0)
-    pos.benchmark_pto(1)
+    # pos = BenchmarkPositivity()
+    # pos.benchmark_pto(0)
+    # pos.benchmark_pto(1)
```
AleCandido, line 307: If possible, consider going in the direction of EKO's [`lha.py`](https://github.com/NNPDF/eko/blob/master/benchmarks/lha_paper_bench.py). I.e. use PyTest markers, instead of commented lines (such that we limit this kind of diffs)

src/yadism/observable_name.py
```diff
@@ -50,6 +50,13 @@ def name(self):
         """joint name"""
         return self.kind + "_" + self.flavor
 
+    @property
+    def is_parity_violating(self):
```
AleCandido, line 54: Not a strong opinion, but for these names I usually prefer dropping the `is_` for simplicity.
On one side it is more or less the same, on the other it is slightly more consistent, because it is a property (the `is_something()` is more consistent for a function, since it is an action, but a property is supposed to be a pseudo-attribute, so it should pretend to be an entity, not an action).

src/yadism/coefficient_functions/light/nlo/g1.py
```diff
@@ -0,0 +1,23 @@
...
+# F2LIGHT_NS
+ns_delta = f2.ns_delta
+ns_omx = f2.ns_omx
+ns_logomx = f2.ns_logomx
+
+
+@nb.njit("f8(f8,f8[:])", cache=True)
+def gluon_reg(z, args):
+    nf = args[0]
+    return 4.0 * nf * TR * ((2.0 * z - 1.0) * np.log((1.0 - z) / z) - 4.0 * z + 3.0)
```
AleCandido, line 23: I'm not sure about the origin of this `4.0` (the one in front of `nf`), but sometimes, to make their role more explicit, we named these factors after their role (i.e. `as_norm`, in those cases - and being NLO it is a likely candidate even fro this one).

src/yadism/coefficient_functions/light/kernels.py
```diff
@@ -77,7 +95,7 @@ def generate(esf, nf, skip_heavylight=False):
     return [ns, g, s]
 
 
-def nc_weights(coupling_constants, Q2, kind, nf, skip_heavylight=False):
+def nc_weights(coupling_constants, Q2, kind, nf, is_pv, skip_heavylight=False):
```
AleCandido, line 98: In principle, being parity preserving or parity violating is a property of the `kind`, so you're passing twice this information, and you may avoid the extra argument.

In order to make it simpler, consider to lift the `is_parity_violating()` function from an `ObservableName` property to a standalone function on kinds (i.e. on `str`s), and use this latter function in the current property implementation.
Consequently, you could use the function on kinds even in here, without duplication.

src/yadmark/benchmark/external/apfel_utils.py
```diff
@@ -24,10 +24,18 @@ def load_apfel(theory, observables, pdf="ToyLH", use_external_grid=False):
...
+    if is_polarized.size != 1:
+        raise ValueError(
+            "Apfel can't compute polarized and unpolarized observables at the same time."
+        )
+
     # set DIS params
     apfel.SetProcessDIS(observables.get("prDIS", "EM"))
     apfel.SetPropagatorCorrection(observables.get("PropagatorCorrection", 0))
     apfel.SetPolarizationDIS(observables.get("PolarizationDIS", 0))
+    apfel.SetPolarizedEvolution(int(is_polarized))
```
AleCandido, line 38:
```suggestion
    apfel.SetPolarizedEvolution(0 if not is_polarized else 1)
```
I know it's the exact same, but it is not an excellent practice to rely on it. However, this is purely a matter of taste, even more because this behavior [is documented](https://docs.python.org/3/library/stdtypes.html#boolean-values). So, feel free to ignore this comment.

AleCandido (commented):
I haven't reviewed yet the NNLO numbers (and I won't do it, of course) and TMC.

However, what I saw is perfectly fine, so you could also consider merging :)

src/yadism/coefficient_functions/light/nlo/g1.py
giacomomagni: I'd leave it this way. This has been copied from `apfel` and follows exactly the equation given in the 
original [paper](https://reader.elsevier.com/reader/sd/pii/055032139490538X?token=9FDADFCED77D43FABB6A22975BB951E40B72D9B7321ED3861C912B9AE144B37FB22C37E30DD888E12AE1203B1681BBF9&originRegion=eu-west-1&originCreation=20230517063035)

benchmarks/runners/apfel_.py
giacomomagni: sure, let's keep this in mind; maybe we can do that in another PR, as we might want to introduce a similar workflow also here....

src/yadism/coefficient_functions/light/kernels.py
giacomomagni: Actually I now notice that `kind` is not used anymore in this function. So we can actually drop it and it's even simpler.

Radonirinaunimi:
> Part of the breaking tests are breaking because we do not support py3.7 any longer, at least in EKO (and consequently also in yadism, at least because of the dependency).
> 
> It should be sufficient to upgrade the version range of _regression_ to be the same of _unit tests_.
> 
> However, even py3.8-9 are failing, and here the workflow complains about a missing `requirements.txt` file. Should we disregard these tests at all (then deactivating), solve here, or solve in a further PR (i.e. open now an issue)?

I was indeed wondering about this, but honestly I think that doing it in another PR would be the best course of actions, if not for the git history.

src/yadism/coefficient_functions/light/kernels.py
giacomomagni: solved in 28b50af3222fa7560d648053cfb5fc081adfdf2b

src/yadmark/benchmark/external/apfel_utils.py
Radonirinaunimi: And also exaplained [here](https://peps.python.org/pep-0285/). So as long as `bool` inherits from `int` this mapping is perfectly fine, so I'd leave it like this.

src/yadism/observable_name.py
Radonirinaunimi: This, I do not also have a strong opinion, but at least doing it this way is consistent with what are currently there:

https://github.com/NNPDF/yadism/blob/0e0dc11ad9e959ba17f380ace38d21ff28187fd3/src/yadism/observable_name.py#L113-L114

https://github.com/NNPDF/yadism/blob/0e0dc11ad9e959ba17f380ace38d21ff28187fd3/src/yadism/observable_name.py#L125-L126

https://github.com/NNPDF/yadism/blob/0e0dc11ad9e959ba17f380ace38d21ff28187fd3/src/yadism/observable_name.py#L130-L131

https://github.com/NNPDF/yadism/blob/0e0dc11ad9e959ba17f380ace38d21ff28187fd3/src/yadism/observable_name.py#L135-L136

https://github.com/NNPDF/yadism/blob/0e0dc11ad9e959ba17f380ace38d21ff28187fd3/src/yadism/observable_name.py#L140-L141

So, in order to be consistent everywhere we'd also need to replace the rests, which probably should better be done in a different PR.

src/yadism/coefficient_functions/light/kernels.py
AleCandido: So much the better :)

felixhekhorn (approved):

AleCandido:
> I was indeed wondering about this, but honestly I think that doing it in another PR would be the best course of actions, if not for the git history.

The main reason why I pointed it out, it's exactly that those workflows are not running in `master`, so they are being reintroduced by this PR.

felixhekhorn:
> > I was indeed wondering about this, but honestly I think that doing it in another PR would be the best course of actions, if not for the git history.
> 
> The main reason why I pointed it out, it's exactly that those workflows are not running in `master`, so they are being reintroduced by this PR.

Well I deactivated them (as said in #190 ) in master because I realised here they were failing and they were stupid - just merge master here (dropping the workflow)

src/yadmark/benchmark/external/apfel_utils.py
Radonirinaunimi: Okay, but out of curiosity, why do you think this specific casting is not suggested?

Radonirinaunimi:
@AleCandido Are you happy for this to be merged now?

Radonirinaunimi:
> About TMC, I have the feeling there might be too much copy-pasting around, but in general (it has been simply passed a lot of time since I took a look at it).
> 
> I would not block this PR certainly because of that, and other refactors are better coupled to numerical FONLL implementation, rather than polarized.

Ok, I'm merging this now.

As for the TMC, apart from the bug in $F_3$ that has already been there, there isn't that much. There is only the `_k` functions implemented in the base class, which currently could be moved outside as they are only used by $g_1$, but when $g_2$ will be implemented in the future these functions will have to be in the base class.

PR #174, opened 2023-02-27
add n3lo ren sv

giacomomagni:
N3LO renomalization SV.

### DIFF
--- a/src/yadism/esf/scale_variations.py
+++ b/src/yadism/esf/scale_variations.py
@@ -126,7 +126,13 @@ class ScaleVariations:
         dict :
             map with `(target, lnf2r, src) -> np.ndarray`
         """
-        ren_coeffs = {(2, 1, 1): -beta.beta_qcd_as2(nf)}
+        beta0 = beta.beta_qcd_as2(nf)
+        ren_coeffs = {
+            (2, 1, 1): -beta0,
+            (3, 1, 2): - 2 *beta0,
+            (3, 1, 1): - beta.beta_qcd_as3(nf),
+            (3, 2, 1): + beta0 ** 2
+        }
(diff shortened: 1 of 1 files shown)


giacomomagni:
Implementing the full SV scheme C at N3LO would require computing a few non trivial convolutions with spilling functions, 
would this work only for renormalisation SV ?

giacomomagni:
> Regarding fact sv, I had a second thought and discussed with @felixhekhorn: since we need in any case a separate integration procedure (we have to integrate a matrix), and the outcome is essentially an EKO, then maybe we will just integrate an EKO, i.e. numerical Mellin invert the N-space expression, rather than convoluting the x-space elements.
> 
> This would allow us to maintain a single set of splitting functions/anomalous dimensions, and it would save the analytic convolutions (by virtue of Mellin).

Okay I agree this can be advantageous for Fact SV at N3LO, I'll add it to the relative issue. 
Regarding Ren SV is this change sufficient?

AleCandido:
> Regarding Ren SV is this change sufficient?

It should be enough, given that the coefficients are the correct ones. About this, they should be the exact same used to upgrade grids for ren sv by @andreab1997 in Pineko, thus I'd suggest he should review this.

andreab1997:
> > Regarding Ren SV is this change sufficient?
> 
> It should be enough, given that the coefficients are the correct ones. About this, they should be the exact same used to upgrade grids for ren sv by @andreab1997 in Pineko (for the specific case of DIS), thus I'd suggest he should review this.

Yes they should be the same but note that I still have to implement N3LO in pineko. Anyway I need to use this expression so let me derive it again and review this PR

andreab1997:
Ok it took me a while to understand the notation but now it is clear and I can confirm it should be correct.

andreab1997 (approved):

PR #178, opened 2023-03-05
Fix F_heavy in `nf = ihq - 2`  

giacomomagni:
This PR is for fixing `F_bottom` in `nf=3` or similarly `F_top` in `nf=3,4`. 
I tested only `EM` and now it agrees perfectly.
If you can confirm this is working also for `NC` (and there are no other changes needed) we can merge it

### DIFF
--- a/src/yadism/coefficient_functions/heavy/kernels.py
+++ b/src/yadism/coefficient_functions/heavy/kernels.py
@@ -19,6 +19,8 @@ def generate(esf, nf, ihq):
             kinematic point
         nf : int
             number of light flavors
+        ihq : int
+            quark flavor to activate
 
     Returns
     -------
@@ -41,7 +43,7 @@ def generate(esf, nf, ihq):
         # F3 is a non-singlet quantity and hence has neither gluon nor singlet-like contributions
         if kind == "F3":
             return ()
-        weights = nc_weights(esf.info.coupling_constants, esf.Q2, kind, nf)
+        weights = nc_weights(esf.info.coupling_constants, esf.Q2, kind, ihq)
         gVV = kernels.Kernel(weights["gVV"], pcs.GluonVV(esf, nf, m2hq=m2hq))
         gAA = kernels.Kernel(weights["gAA"], pcs.GluonAA(esf, nf, m2hq=m2hq))
         sVV = kernels.Kernel(weights["sVV"], pcs.SingletVV(esf, nf, m2hq=m2hq))
@@ -81,7 +83,7 @@ def generate_missing(esf, nf, ihq, icoupl=None):
     return (kernels.Kernel(weights["ns"], pcs.NonSinglet(esf, nf, m2hq=m2hq)),)
 
 
-def nc_weights(coupling_constants, Q2, kind, nf):
+def nc_weights(coupling_constants, Q2, kind, ihq):
     """
     Compute heavy NC weights.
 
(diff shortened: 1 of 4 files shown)


AleCandido:
Thanks, this transition happened inconsistently, from an era in which `ihq` was blocked to `nf + 1`. It has more or less being done in #124, but directly at the level of GM-VFNS, so the individual building blocks were not carefully checked.

Since this was an undetected error, it would be ideal to have some unit tests, that would be broken in current `master`, but passing after this PR (if possible, linked to some kind of invariant, relevant property, rather than testing numbers).

tests/yadism/cf/test_nc_kernels.py
```diff
@@ -54,10 +54,9 @@ def mkpv(nf, w):  # pv = parity violating
 def check(ps, w):
     assert len(w) == len(ps)
     for e, k in zip(ps, w):
-        assert pytest.approx(e) == k.partons
+        assert e == k.partons
```
AleCandido, line 57: Are you sure you don't need `approx`? If you are comparing `float`s is always required, you should omit only for `int`s

giacomomagni: Most likely tests which were comparing floats are still skipped.
I wasn't sure how to restore also the `fonll_diff` ones.

AleCandido: About FONLL diff I'd not worry that much, since #175 should remove it.

andreab1997:
So is this completed?

felixhekhorn:
If we can let's fix the CC case as well!
- for `heavy` I believe we have to change these lines https://github.com/NNPDF/yadism/blob/937d7718819408c8a532be8dbf5946b73eb622e1/src/yadism/coefficient_functions/heavy/kernels.py#L35-L37 - specifically, the `cc_mask` (4th arg) should use `ihq` (as the TODO says) - actually it should be `ihq-1` (as the usual translation should be `ihq = nf+1`); to make an example for bottom `ihq` is `5`, right? and so we need the fourth element of that list, I think - do you agree?
- for `fonll` this is actually already implemented this way: https://github.com/NNPDF/yadism/blob/a3a308e9f7d46e05a8a42bcc88d71c8135fbac24/src/yadism/coefficient_functions/fonll/kernels.py#L187

felixhekhorn:
> So is this completed?

If someone can double check my commit, yes

felixhekhorn (approved):

andreab1997:
> > So is this completed?
> 
> If someone can double check my commit, yes

It seems reasonable to me

src/yadism/coefficient_functions/heavy/kernels.py
```diff
@@ -1,37 +1,53 @@
...
+
     """
     kind = esf.info.obs_name.kind
     pcs = import_pc_module(kind, esf.process)
     m2hq = esf.info.m2hq[ihq - 4]
     if esf.process == "CC":
-        # TODO: use ihq for couplings
         w = kernels.cc_weights(
-            esf.info.coupling_constants, esf.Q2, kind, br.quark_names[nf], nf
+            esf.info.coupling_constants, esf.Q2, kind, br.quark_names[ihq - 1], nf
```
giacomomagni, line 50: why do you need to pass `ihq` here?  If you have `F_bottom` in `nf=3` only `g,u,d,s` (and relative - pids) are around right?

felixhekhorn: yes - because that means with the bottom mask ([remember the masks in CC](https://yadism.readthedocs.io/en/latest/theory/fns.html)), we have $V_{ub}$ in game and so `F_bottom` can be non zero (exactly due to $u$-PDF - and of course at NLO through $g$)

giacomomagni: okay I see thanks for the clarification.

Issue #176, opened 2023-02-28
Minus sign missing in k_qq_loc


giacomomagni:
While trying to benchmarking `DIS_TP` I might have found a global minus sign missing in this function:

https://github.com/NNPDF/yadism/blob/e3eab124b25083c32289134c8204a8bf0f4b33b8/src/yadism/coefficient_functions/fonll/partonic_channel.py#L108

here you have a test checking the convolution of `K_qq` and a test function, both with Yadism and Eko:

```
import numpy as np
from scipy.integrate import quad
from eko.mellin import Path
from ekore.harmonics import compute_cache
from ekore.operator_matrix_elements.unpolarized.space_like import as2
from numpy.testing import assert_allclose


from yadism.coefficient_functions.fonll.partonic_channel import (
    PdfMatchingNNLLNonSinglet,
)
from test_pc_general import MockESF


nf = 4

# a simple test function and its mellin transform
def f(x):
    return 1

def mellin_f(n):
    return 1/(n)

# x-space convolution
def yad_convolute(x, q):
    esf = MockESF(x, q**2)
    match = PdfMatchingNNLLNonSinglet(esf, nf, m2hq=q**2).NNLO()
    loc = match.loc(x, match.args["loc"]) * f(x)
    sing = quad(
        lambda z: match.sing(z, match.args["sing"]) * (f(x / z) / z - f(x)),
        x,
        1,
    )[0]
    return sing + loc

# n-space convolution
def inverse_mellin(x):
    def quad_ker_talbot(u, func):
        path = Path(u, np.log(x), True)
        sx = compute_cache(path.n, 3, True)
        sx = [np.array(s) for s in sx]
        integrand = path.prefactor * x ** (-path.n) * path.jac
        gamma = func(path.n, sx, L=0) * mellin_f(path.n)
        return np.real(gamma * integrand)

    return quad(
        lambda u: quad_ker_talbot(u, as2.A_qq_ns),
        0.5,
        1.0,
        epsabs=1e-12,
        epsrel=1e-6,
        limit=200,
        full_output=1,
    )[0]


class Test_Matching_qq:
    xs = [0.0001, 0.001, 0.01, 0.1, 0.2, 0.456, 0.7]
    Qs = [5, 10, 20]

    def test_nnlo(self):
        my = []
        eko = []
        for q in self.Qs:
            for x in self.xs:
                my.append(yad_convolute(x, q))
                eko.append(inverse_mellin(x))
        assert_allclose(my, eko)
[...]

PR #177, opened 2023-03-03
Really fix FONLL matching away from mass

felixhekhorn:
In the end #170 was quite pointless since the crucial fix didn't survive ...

- in https://github.com/NNPDF/yadism/pull/170/commits/7a81fefe599f5f3afba87565d0c31c3091847631 the correct behaviour is implemented, but with a wrong name (`mu2hq = esf.info.m2hq[ihq - 4]`)
- in https://github.com/NNPDF/yadism/pull/170/commits/f3fbff3132c048d8836ac9e5c465d7efb08934d5 this correction got (most likely unintended) reverted (`m2hq = esf.info.threshold.area_walls[ihq - 3]`) by @AleCandido 

This PR should restore the good behaviour, i.e. coefficient functions should not depend on the matching scale (but that dependency is fully managed by EKO)

@andreab1997 @giacomomagni did I forgot an appearance? @andreab1997 are your grids before or after that commit? :see_no_evil: 

let's close this ASAP - Note to future developers: let's write a unit test for this (should be simple)

PR #179, opened 2023-03-06
Missing minus sign in local part of K_qq matching

giacomomagni:
Address #176


andreab1997:
If we confirm this is a bug and we merge this, I can start producing the new DIS grids (which have to be recomputed also because of #177 ). Or do you know of other bugs?

felixhekhorn:
I added the check to the Buza representation - and unlike in https://github.com/NNPDF/yadism/blob/13d00c92501bc765901b308850ef7729a34c2f31/tests/yadism/cf/test_fonll_pc.py#L86  there is no limit here (so default `rtol=1e-7`)

felixhekhorn (approved):

giacomomagni:
Okay thanks for double testing @felixhekhorn!

andreab1997:
Let's wait for #178 then

PR #180, opened 2023-03-08
add g4_reg nc

RoyStegeman:
Includes polarized structure function $g_4$ at NLO

Reference for NNLO (eq. 19): https://arxiv.org/pdf/2210.12014.pdf


benchmarks/runners/apfel_.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
```
Radonirinaunimi, line 1: It looks like the formatter from the pre-commit hooks deleted the enconding (?). This is not that much needed since python>=3 but maybe for the consistency with the other files we should keep it.

src/yadism/coefficient_functions/light/g4_nc.py
```diff
@@ -0,0 +1,51 @@
+from ..partonic_channel import RSL
+from . import nlo
+from . import partonic_channel as pc
+
+
+class NonSinglet(pc.LightBase):
+    @staticmethod
+    def LO():
+        """
+        |ref| APFEL
```
Radonirinaunimi, line 10: You can actually cite Eq. (4) of hep-ph/9412255. In this case it 'll be more apparent that it is indeed a delta term.

src/yadism/coefficient_functions/light/g4_nc.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+
+        # leading order is just a delta function
+        return RSL.from_delta(1.0)
+
+    @staticmethod
+    def NLO():
+        """
+        |ref| hep-ph/9412255
```
Radonirinaunimi, line 19: You could already point to the reference in the bibliography.

https://github.com/NNPDF/yadism/blob/030acf43e6fe2001269b3761c92e639b1fc1a9c7/src/yadism/coefficient_functions/light/g1_nc.py#L21

src/yadism/coefficient_functions/light/g4_nc.py
```diff
@@ -0,0 +1,51 @@
...
+
+        Note
+        ----
+        2 * n_f is coming from momentum sum rule q_i -> {q_i, g} but
+        g -> {g, q_i, \bar{q_i} forall i}, so the 2 * n_f is needed
+        to compensate for all the number of flavours plus antiflavours
+        in which the gluon can go.
+        """
+
+        return RSL(nlo.g1.gluon_reg, args=[self.nf])
```
Radonirinaunimi, line 43: I think that the `g1_gluon`-contribution is zero (same as `g1_gluon`). But I am not sure if returning zero in `nlo.g4.gluon_reg`  would be the same as calling `pc.EmptyPartonicChannel`. cc @felixhekhorn, @AleCandido 

```suggestion
class Gluon(pc.EmptyPartonicChannel):
    pass
```

benchmarks/runners/apfel_.py
AleCandido: We're just incrementally removing them, with `pre-commit`, in order to avoid a large diff.

As the formatter explains, that is a bunch of useless characters by now, so we will remove. But if you do all at once, you generate a large diff with a lot of files changed. So there are two options: either we do it all at once in a single commit, or we just update files changed. `pre-commit` is already doing by default the second, so we thought there was no need for the first :)

src/yadism/coefficient_functions/light/g4_nc.py
AleCandido: You can look at the definition: `EmptyPartonicChannel` is just a marker, but adds nothing to the base class:
https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/src/yadism/coefficient_functions/partonic_channel.py#L171-L173

and the base class definitions for the orders is pretty simple:
https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/src/yadism/coefficient_functions/partonic_channel.py#L154-L160

benchmarks/runners/apfel_.py
Radonirinaunimi: I agree then that removing it for the changed files would be fine, and if at some point one would like to remove it in all the files then one can do so in a single commit (the number of files changed then will be much smaller).

Please ignore the above comment then!

src/yadism/coefficient_functions/light/nlo/g4.py
```diff
@@ -0,0 +1,26 @@
+import numba as nb
+import numpy as np
+from eko.constants import CF, TR
+
+from . import f2
+
+
+@nb.njit("f8(f8,f8[:])", cache=True)
+def ns_reg(z, _args):
```
giacomomagni, line 9: this piece should be the same as `f2.ns_reg`

RoyStegeman: Where do you get that from? I got this expression from eq.17 here: https://arxiv.org/pdf/hep-ph/9412255.pdf, which also seems to be what is implemented in apfel.

giacomomagni: Looking [here](https://arxiv.org/pdf/2210.12014.pdf), you can see that eq 19 holds up to NNLO. 
I tried with Mathematica and the expressions are indeed the same.

RoyStegeman: Indeed also [here](https://arxiv.org/pdf/hep-ph/9509236.pdf) seems to agree with those relations. it's also the same as what is currently implemented but I can indeed directly import the function from f2.

src/yadmark/benchmark/external/apfel_utils.py
```diff
@@ -25,7 +25,7 @@ def load_apfel(theory, observables, pdf="ToyLH", use_external_grid=False):
         theory, observables, pdf, use_external_grid=use_external_grid
     )
 
-    is_polarized = ["g1" in obs_name for obs_name in observables["observables"]]
+    is_polarized = ["g" in obs_name for obs_name in observables["observables"]]
```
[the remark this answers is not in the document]
AleCandido: `obs_name.startswith("g")`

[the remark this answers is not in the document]
AleCandido: `obs_name.startswith("g")`

src/yadmark/benchmark/external/apfel_utils.py
```diff
@@ -36,6 +36,7 @@ def load_apfel(theory, observables, pdf="ToyLH", use_external_grid=False):
     apfel.SetPropagatorCorrection(observables.get("PropagatorCorrection", 0))
     apfel.SetPolarizationDIS(observables.get("PolarizationDIS", 0))
     apfel.SetPolarizedEvolution(int(is_polarized))
+    # apfel.SetPolarizedEvolution(observables.get("PolarizationDIS", 0))
```
giacomomagni, line 39:
```suggestion
```

Radonirinaunimi:
LGTM!

Just to mention that now all the references to the de Florian paper should be updated:
https://github.com/NNPDF/yadism/blob/44341a3dad7451cb57487c32a5a51924566d7e42/src/yadism/coefficient_functions/light/g1_nc.py#L22

https://github.com/NNPDF/yadism/blob/44341a3dad7451cb57487c32a5a51924566d7e42/src/yadism/coefficient_functions/light/g1_nc.py#L37

RoyStegeman:
That's the nlo paper, this is the nnlo paper.

Radonirinaunimi:
> That's the nlo paper, this is the nnlo paper.
> 
> Or do you mean replace the nlo with the nnlo paper, I don't think that matters very much if both are correct.

Yes, I meant replace. But as you say it doesn't matter that much, at least unless we use the equivalence relation everywhere.

So, please feel free to merge this into #173.

PR #181, opened 2023-03-16
Apfel++ benchmarks

giacomomagni:
This PR is to introduce APFEL++ in yadmark. 
This will be needed to benchmark polarized SF.


giacomomagni:
@Radonirinaunimi Yadmark is running, 
but numbers are obviously crazy, so there is still some work to to in `apfelpy_utils`, it would be great if you can take a look as well.

Note that while adding this I noticed that we still have to fix all the couplings for polarised SF in case of CC and NC.

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -0,0 +1,168 @@
...
+            "gL": init.InitializegLNCObjectsZM,
+            "g4": init.Initializeg4NCObjectsZM,
+        }
+
+    # TODO: do you need to disable evolution also here?
+    # disable APFEL++ evolution: we are interested in the pure DIS part
+    dglapobj = ap.initializers.InitializeDglapObjectsQCD(xgrid, thrs)
+    # Construct the DGLAP objects
+    evolvedPDFs = ap.builders.BuildDglap(
+        dglapobj, ap.utilities.LHToyPDFs, theory["Q0"], pto, alphas.Evaluate
```
giacomomagni, line 87: Here I think we should pass `pdf.xfxQ`, but somehow Apfel++ was complaining about the banana object..

Radonirinaunimi:
> @Radonirinaunimi Yadmark is running, but numbers are obviously crazy, so there is still some work to to in `apfelpy_utils`, it would be great if you can take a look as well (I will not commit anything else today or tomorrow so feel free to push).
> 
> Note that while adding this I noticed that we still have to fix all the couplings for polarised SF in case of CC and NC adjusting the parity violating terms.

Thanks a lot @giacomomagni! I can indeed take it from here.

benchmarks/runners/apfelpy_.py
AleCandido, line 1: Trivial question: why `apfelpy_py.py` (with 3 `py`s), instead of `apfelxx_.py`? (just the closest option to the existing `apfel_.py`)

giacomomagni: As you wish the python binding is called `apfelpy`

Radonirinaunimi: I would also call it `apfelxx_.py`. This is at least consistent with `apfel_.py`.

AleCandido: > Both are fine for me, the python binding of Apfel++ is called `apfelpy`

Ok, fine, then your choice makes sense.

I wonder why Valerio called it `apfelpy`, but given this name, both choices make sense within yadism. I do not have a personal preference...

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # iterate over input kinematics
+        for kin in kinematics:
+            Q2 = kin["Q2"]
+            x = kin["x"]
+
+            # Construct the DGLAP objects
+            evolvedPDFs = ap.builders.BuildDglap(
+                dglapobj,
+                ap.utilities.LHToyPDFs,
```
giacomomagni, line 131: Here we still need to pass the generic pdf...

src/yadism/coefficient_functions/fonll/kernels.py
```diff
@@ -194,7 +194,7 @@ def generate_heavy_diff(esf, nl, pto_evol):
         asy_weights = heavy.kernels.nc_weights(
             esf.info.coupling_constants, esf.Q2, kind, nl, ihq
         )
-        if kind != "F3":
+        if kind not in ["F3", "gL", "g4"]:
```
felixhekhorn, line 197: How about adding a (static) function in obsname module `is_pv` or `is_parity_violationg` or `parity_violation` or similar which return exactly this if condition?

Radonirinaunimi: Yes, I agree this would be ideal.

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -0,0 +1,231 @@
...
+NQ = 50
+QMin = 1
+QMax = 200
+
+# Map observables names to APFEL++ methods
+MAP_HEAVINESS = {
+    "charm": 4,
+    "bottom": 5,
+    "top": 6,
+    "light": 0, # TODO: Check combination
```
Radonirinaunimi, line 16: Here we'd want to also allow for a compound matching given how the light contributions are computed in Apfel++:

https://github.com/vbertone/apfelxx/blob/4ebe6952888f238457fb0b72cdfec5963b26b553/pywrap/structurefunction_nc_test.py#L51.

src/yadism/coefficient_functions/fonll/kernels.py
Radonirinaunimi: This is done now! Some functions had to be extended but I think this will not break backward compatibility.

src/yadmark/benchmark/external/apfelpy_utils.py
giacomomagni: yes regarding this if summing only on 1,2,3 is easy enough we can do it otherwise we look only at total. 
In the end we can't accommodate all the definitions which are around. We know polarised is not making any difference in this respect (ie heaviness)

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -0,0 +1,312 @@
...
+            thrs,
+        )
+    elif "charm" in obs_name:
+        tab_sf = ap.TabulateObjectD(sfobj[4].Evaluate, nq, qmin, qmax, 3, thrs)
+    elif "bottom" in obs_name:
+        tab_sf = ap.TabulateObjectD(sfobj[5].Evaluate, nq, qmin, qmax, 3, thrs)
+    return tab_sf
+
+
+def tabulate_nc(ap, obs_name, sfobj, nq, qmin, qmax, thrs):
```
giacomomagni, line 112:
```suggestion
def tabulate_cc(ap, obs_name, sfobj, nq, qmin, qmax, thrs):
```

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -0,0 +1,312 @@
...
+            raise ValueError("Yadism cannot compute polarised CC yet.")
+    else:
+        coupling = _fBq
+        if "F3" in obs_name or "gL" in obs_name or "g4" in obs_name:
+            coupling = _fDq
+
+    return coupling
+
+
+def tabulate_ncc(ap, obs_name, sfobj, nq, qmin, qmax, thrs):
```
giacomomagni, line 69:
```suggestion
def tabulate_nc(ap, obs_name, sfobj, nq, qmin, qmax, thrs):
```

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -0,0 +1,312 @@
...
+        return ap.utilities.ParityViolatingElectroWeakCharges(Q, False, pids)
+
+    # CKM matrix elements
+    def _fCKM(Q):
+        return ap.constants.CKM2
+
+    if proc_type == "CC":
+        coupling = _fCKM
+        if obs_name.startswith("g"):
+            raise ValueError("Yadism cannot compute polarised CC yet.")
```
giacomomagni, line 60:
```suggestion
            raise ValueError("Apfel++ cannot compute polarised CC yet.")
```
Let's check here only things which are not implemented in Apfel++. 
We are planning to implement them soon, but not sure if Valerio is planning to add them as well. 
Yadism input checks should go here https://github.com/NNPDF/yadism/blob/master/src/yadism/input/

src/yadmark/benchmark/external/apfelpy_utils.py
Radonirinaunimi: Ah, well spotted! This should indeed be `_cc`.

src/yadmark/benchmark/external/apfelpy_utils.py
Radonirinaunimi: But this is not actually fully correct, right! Since `EM` will also pass there.

AleCandido: EM is considered to be a subset of NC

Radonirinaunimi: So you're saying that calling it `_nc` would be fine? I was trying to call it Non-Charged-Current (`_ncc`) but I will also be happy with that.

src/yadmark/benchmark/external/apfelpy_utils.py
Radonirinaunimi: @giacomomagni I think this has now been more or less fixed. What I don't fully understand though is why these `pids` have to be passed to the coupling functions? In some of Valerio's examples, this entry is not needed.

https://github.com/NNPDF/yadism/blob/f64423182627df92e4035d7e7a31316b97fa5968/src/yadmark/benchmark/external/apfelpy_utils.py#L42-L46

giacomomagni: I think this function returns a list of couplings for each pid. If not given all couplings are returned, no?

Radonirinaunimi: But to be consistent with Valerio's examples, should we remove this entry?

Radonirinaunimi: > fine, I think in Valerio's examples he's always computing total that's why he is not specifying any pid.

Wait, he is also computing [light, charm, bottom structure functions](https://github.com/vbertone/apfelxx/blob/00238cd9d5711e6ac16e88dad1e17ba08761c32b/pywrap/structurefunction_nc_pol_test.py#L50-L53).

giacomomagni: I also think the light combination will not match due to the definition. If I recall correct we don't match apfel light definition ( @felixhekhorn is this correct? ), so most likely we don't also here.

giacomomagni: > Wait, he is also computing [light, charm, bottom structure functions](https://github.com/vbertone/apfelxx/blob/00238cd9d5711e6ac16e88dad1e17ba08761c32b/pywrap/structurefunction_nc_pol_test.py#L50-L53).

But he is always doing total and the other... Never only charm or bottom

AleCandido: "light" is in a complex state, because I believe its meaning might depend on the FNS (I should check). Originally, we were defining "light" as zero-mass, but this would not work for FONLL, so I'm almost sure "light" in FONLL means another thing.

However, this hopefully will simplify after the implementation of "numerical FONLL", that for yadism means essentially drop FONLL implementation, and implement just the ingredients.

giacomomagni: Indeed I believe we can just drop light here, 
as said we haven't changed anything in that respect with polarised.

benchmarks/runners/apfelpy_.py
```diff
@@ -88,7 +84,7 @@ def benchmark_polarized(self, pto):
         self.run(
             self.theory_updates_zm(pto),
             self.obs_updates_pol(),
-            ["ToyLH"],
+            ["ToyLH_polarized"],
```
Radonirinaunimi: Using `ToyLH_polarized` significantly worsen the agreement; at LO the difference can be up to 12%. This also happens when benchmarking against Apfel.

benchmarks/runners/apfelpy_.py
```diff
@@ -0,0 +1,94 @@
...
+            self.theory_updates_zm(pto),
+            self.obs_updates_zm(),
+            ["ToyLH"],
+        )
+
+    def benchmark_polarized(self, pto):
+        self.run(
+            self.theory_updates_zm(pto),
+            self.obs_updates_pol(),
+            ["ToyLH_polarized"],
```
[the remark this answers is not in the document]
giacomomagni: it seems that using `ToyLH_polarized` enehanced the difference in `g1` at NNLO, which was barely visible before.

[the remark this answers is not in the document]
felixhekhorn: can you break the difference down by channel? (using genpdf) how about using the good old $x(1-x)$?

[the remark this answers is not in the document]
giacomomagni: we can try, the difference should originate from the Non Singlet coefficient.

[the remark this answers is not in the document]
giacomomagni: here you have. At NNLO using `-p toy` and using a finer Q grid (`nQ=1000`) for Apfel++: 

`gonly`: 

```
 ────────── 
  g1_total  
 ────────── 
           x            Q2        yadism  yadism_error       APFEL++  percent_error
0   0.000012     10.000000 -2.191937e-02  7.779414e-09 -2.189800e-02       0.097604
1   0.000049     10.000000 -1.882649e-02  2.396574e-09 -1.880779e-02       0.099409
2   0.000204     10.000000 -1.671502e-02  3.147698e-09 -1.669876e-02       0.097360
3   0.000853     10.000000 -1.669766e-02  2.679573e-09 -1.668385e-02       0.082757
4   0.003562     10.000000 -2.029463e-02  2.417312e-09 -2.028393e-02       0.052760
5   0.014874     10.000000 -2.775710e-02  1.967232e-09 -2.775080e-02       0.022690
6   0.062102     10.000000 -3.306795e-02  1.190518e-09 -3.307135e-02      -0.010280
7   0.194737     10.000000 -1.990439e-02  5.202715e-10 -1.990439e-02       0.000014
8   0.336842     10.000000 -7.652458e-03  1.032559e-10 -7.652416e-03       0.000550
9   0.478947     10.000000 -2.121734e-03  4.312387e-11 -2.121466e-03       0.012611
10  0.621053     10.000000 -3.720199e-04  6.460554e-12 -3.723791e-04      -0.096471
11  0.763158     10.000000 -2.783076e-05  4.332267e-13 -2.746872e-05       1.317991
12  0.905263     10.000000 -1.239917e-07  7.448662e-15 -9.908782e-08      25.133165
13  0.001000      3.000000 -1.570717e-02  3.086463e-09 -1.572291e-02      -0.100142
14  0.001000      7.388431 -1.854799e-02  3.559920e-09 -1.856698e-02      -0.102262
15  0.001000     18.196305 -1.423482e-02  2.678166e-09 -1.424964e-02      -0.104038
16  0.001000     44.814047 -1.284337e-02  2.378096e-09 -1.285694e-02      -0.105483
17  0.001000    110.368500 -1.083159e-02  1.976833e-09 -1.084318e-02      -0.106816
18  0.001000    271.816686 -9.335745e-03  1.681518e-09 -9.345855e-03      -0.108177
19  0.001000    669.432950 -8.210687e-03  1.461093e-09 -8.219712e-03      -0.109805
20  0.001000   1648.686404 -7.421305e-03  1.305974e-09 -7.429647e-03      -0.112284
21  0.001000   4060.401956 -7.062689e-03  1.230100e-09 -7.070925e-03      -0.116467
22  0.001000  10000.000000 -7.219567e-03  1.245420e-09 -7.228399e-03      -0.122188
```

`lonly` ie light quarks only: 

```
 ────────── 
  g1_total  
 ────────── 
           x            Q2    yadism  yadism_error   APFEL++  percent_error
0   0.000012     10.000000 -0.000291 -2.251389e-11 -0.000287       1.305583
1   0.000049     10.000000 -0.000344 -4.627033e-11 -0.000340       1.095596
2   0.000204     10.000000 -0.000373 -1.629789e-11 -0.000373       0.057539
3   0.000853     10.000000 -0.000240  4.838951e-12 -0.000254      -5.772452
4   0.003562     10.000000  0.001481  5.130205e-11  0.001418       4.454185
5   0.014874     10.000000  0.012064  1.868573e-10  0.011870       1.632602
6   0.062102     10.000000  0.050398  7.749967e-10  0.049997       0.800746
7   0.194737     10.000000  0.109347  2.073501e-09  0.108995       0.322261
8   0.336842     10.000000  0.126687  5.465003e-09  0.126531       0.123587
9   0.478947     10.000000  0.110273  4.325416e-09  0.110178       0.086218
10  0.621053     10.000000  0.072825  2.807101e-09  0.072871      -0.063313
11  0.763158     10.000000  0.031264  8.198026e-10  0.031163       0.323404
12  0.905263     10.000000  0.004238  9.329867e-11  0.004395      -3.578833
13  0.001000      3.000000  0.000064  4.590512e-12  0.000041      56.256089
14  0.001000      7.388431 -0.000302  9.068681e-12 -0.000322      -6.256600
15  0.001000     18.196305 -0.000009  6.725702e-12 -0.000024     -61.740713
16  0.001000     44.814047  0.000060  4.396884e-12  0.000046      30.844171
17  0.001000    110.368500  0.000175  3.627017e-12  0.000164       6.923428
18  0.001000    271.816686  0.000250  3.047234e-12  0.000241       3.877718
19  0.001000    669.432950  0.000290  2.564404e-12  0.000283       2.767274
20  0.001000   1648.686404  0.000275  2.045430e-12  0.000269       2.471179
21  0.001000   4060.401956  0.000128  1.240913e-12  0.000122       4.640097
22  0.001000  10000.000000 -0.000245 -3.842776e-14 -0.000250      -1.908960
```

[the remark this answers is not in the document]
felixhekhorn: this looks fine to me ...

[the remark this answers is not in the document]
felixhekhorn: so was the problem `nQ`?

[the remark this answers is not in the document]
Radonirinaunimi: > so was the problem `nQ`?

Yes! I am just now investigating if `nQ=1000` is too much or it is enough to use `nQ=200`.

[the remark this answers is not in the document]
giacomomagni: > so was the problem `nQ`?

But only partly, just to be clear the problem is still there but there is not much we can do. 
We are implementing the analytical relation for NS `g1 = F3` while Aplel++ is not. In the bench you see the difference, which is small but not fully invisible.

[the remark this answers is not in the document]
Radonirinaunimi: But this is an analytical relation and I am not sure we'll ever do anything about this.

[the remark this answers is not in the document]
Radonirinaunimi: Also note that now with `nQ=1000`, the agreement for the full NNLO is now ~per-mille level (except for when the absolute results are ~zero).

Radonirinaunimi (approved):

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -0,0 +1,346 @@
...
+NQ = 250
+QMin = 1
+QMax = 200
+
+# Map observables names to APFEL++ methods
+MAP_HEAVINESS = {
+    "charm": 4,
+    "bottom": 5,
+    "top": 6,
+    "light": 0,  # TODO: Check combination
```
felixhekhorn, line 15: what is the status of this todo?

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -0,0 +1,346 @@
...
+    pto = theory["PTO"]
+
+    # Couplings
+    alphas = ap.AlphaQCD(theory["alphas"], theory["Qref"], thrs, pto)
+    alphas = ap.TabulateObject(alphas, 2 * NQ, QMin, QMax, 3)
+
+    # Map Yadism observables to Apfel++ Objects
+    init = ap.initializers
+    if observables["prDIS"] == "CC":
+        # TODO: CC are not running ... CKM matrix error
```
felixhekhorn, line 245: what is the status of this todo?

src/yadmark/benchmark/runner.py
```diff
@@ -92,26 +92,28 @@ def run_external(self, theory, ocard, pdf):
...
-            from .external import (  # pylint:disable=import-error,import-outside-toplevel
-                xspace_bench_utils,
-            )
+            from .external import \
+                xspace_bench_utils  # pylint:disable=import-error,import-outside-toplevel
 
             return xspace_bench_utils.compute_xspace_bench_data(theory, observable, pdf)
 
+        elif self.external.upper() == "APFEL++":
+            from .external import apfelpy_utils
```
felixhekhorn, line 113: please check if and where the pylint fix is needed

felixhekhorn (changes requested):
please fix pre-commit as much as you can

src/yadmark/benchmark/external/apfelpy_utils.py
giacomomagni:
```suggestion
```
This should be fixed.

src/yadmark/benchmark/external/apfelpy_utils.py
giacomomagni: They don't match. So either we leave it as pending either we remove

src/yadmark/benchmark/runner.py
Radonirinaunimi: The `pylint: disable=import-error` is not needed here given that it is instead addressed here:

https://github.com/NNPDF/yadism/blob/4d15c433a55b8622d61db7d869cdc663a0153ab0/src/yadmark/benchmark/external/apfelpy_utils.py#L206

src/yadmark/benchmark/external/apfelpy_utils.py
Radonirinaunimi: Done!

src/yadmark/benchmark/external/apfelpy_utils.py
Radonirinaunimi: Here the best would be to leave it as pending.

Radonirinaunimi:
As for the pre-commit, most of the pydocstyle fails are not related to this PR and would better be addressed in a different PR. 

As for black & isort, they modify all the pylint labels as follows:

```python
           from .external import (  # pylint:disable=import-error,import-outside-toplevel
                apfel_utils,
            )
```
Do we want to introduce these?

felixhekhorn:
> As for the pre-commit, most of the pydocstyle fails are not related to this PR and would better be addressed in a different PR.

agreed

> As for black & isort, they modify all the pylint labels as follows:
> 
> ```python
>            from .external import (  # pylint:disable=import-error,import-outside-toplevel
>                 apfel_utils,
>             )
> ```
> 
> Do we want to introduce these?

if they want so, yes

felixhekhorn (approved):

Issue #182, opened 2023-03-17
Number of active nf in Singlet FONLL


giacomomagni:
While trying to produce the K-factors suggested today I might have found another small hiccup. 
When `skip_heavylight` is used to generate the `nc_weights` the heavy parton is not coupling to the boson 
as correctly said, but still the singlet (and the gluon) charges are divided by the number of light flavors plus the heavy quark `nl+1`. 
I'm not 100% sure this is actually what we want since then only `nl` should be actually active.

So are we sure that here we should have: 
https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/src/yadism/coefficient_functions/light/kernels.py#L91

and not:
```py
    n_pids = len(pids)
    if skip_heavylight:
         n_pids -= 1
    ch_av = tot_ch_sq / n_pids if kind != "F3" else 0.0
```

And similarly: 

https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/src/yadism/coefficient_functions/fonll/kernels.py#L132

and not: 
```py
k = kernels.Kernel(s_w, light_cfs.Singlet(esf, nl)) 
```

Note this will not make any difference up to NNLO where the nf dependency of the Singlet coefficient functions factors out, so for instance for charm thr doing `0.1666 * 4 == 0.222 *3`, but at N3LO this makes a difference. 

@felixhekhorn, @AleCandido, @andreab1997 
Could you please confirm this is the behaviour we want ie:
  $\frac{1}{nl+1} \Sigma_{i}^{nl} e^2_i Singlet(nl+1)$  and not $\frac{1}{nl} \Sigma_{i}^{nl} e^2_i Singlet(nl) $.

giacomomagni:
On a similar point for the heavy SF can you confirm that also here: 

https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/src/yadism/coefficient_functions/kernels.py#L315-L325

we have the desired behaviour. 

I.e the nf parton is not coupling to the boson in singlet diagrams but singlet and gluon are called with nf. 

And not: 

```
 ns_partons[ihq] = w 
 ns_partons[-ihq] = w if kind != "F3" else -w 
 ch_av = w / (nf) if kind != "F3" else 0.0 
 for pid in range(1, nf + 1): 
     s_partons[pid] = ch_av 
[...]

PR #183, opened 2023-03-19
Fix nf dependency of FONLL flight at N3LO

giacomomagni:
Give the discussion in #182 I realised that the N3LO contribution in `Flight`, above the first active FONLL mass, was not reducing correctly to `ZM(nl+1)` with `(nl+1)-th` flavour not coupled with the boson but was doing `ZM(nl) + Singlet(nl+1)`. So I propose the following fix.

The reason of this is mainly due to this cancellation:

https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/src/yadism/coefficient_functions/fonll/kernels.py#L44-L45

which is something that is correct only up to NNLO, since at N3LO the structure of missing diagrams will be more complicated (which of course we don't know and are correctly set to 0)


giacomomagni:
Having this in mind I am also wandering if the `nf` dependency if these terms:

https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/src/yadism/coefficient_functions/light/n3lo/common.py#L13-L26


which are described to [Fig1 and Table2 of this paper](https://arxiv.org/pdf/hep-ph/0504242.pdf), has to be fixed in case of ´Flight´ FONLL. 
Intact the averaged charge we have around should be equal to `nf=nl+1` with the last flavour set to zero only in the numerator. 
This would imply the N3LO coefficient for FONLL depends both on `nl` and `nf`, while for ZM  `nl=nf` always. 

What do you think @AleCandido @felixhekhorn @andreab1997 ??

giacomomagni:
As discussed with @felixhekhorn and @andreab1997 we decided that a proper fix involves: 

* [ ] In FONLL `Flight` above the first active threshold the `color_factors` have to be computed on `nl+1` setting the last charge to 0.
* [ ] Coefficients have to be adjusted for  NC ( luckily `fl11` vanishes for all the CC diagrams :) ) , to include the averaged partonic coupling
* [ ] This PR also contains a fix in `F3_nc` Non Singlet, in which the `fl_02` diagrams should be removed since they belong to Singlet.

src/yadism/coefficient_functions/coupling_constants.py
```diff
@@ -270,6 +270,43 @@ def get_weight(self, pid, Q2, quark_coupling_type, cc_mask=None):
...
+
+        w_ph = self.electric_charge[pid]
+        # pure EM charge
+        if self.obs_config["process"] == "EM":
+            return w_ph
+        # allow Z to be mixed in
+        if self.obs_config["process"] == "NC":
+            w_ZV = self.vectorial_coupling(pid)
+            w_ZA = self.weak_isospin_3[pid]
+            return w_ph + w_ZV + w_ZA
```
giacomomagni, line 307: Did we agreed on sum right, not average?? @andreab1997 @felixhekhorn

felixhekhorn: actually, I wonder about the axial coupling ... a diagram with a single $\gamma_5$ can not survive - so maybe only the vectorial part here ...

andreab1997: Should we keep only the vectorial part then?

AleCandido: > actually, I wonder about the axial coupling ... a diagram with a single γ5 can not survive - so maybe only the vectorial part here ...

Which is the diagram we're talking about?

giacomomagni: [Fig1 and Table2 of this paper](https://arxiv.org/pdf/hep-ph/0504242.pdf)

AleCandido: Ok, sorry, you're right. I was thinking to the cross-section.

But if we're not considering the whole cross-section we can't rule out the axial contribution. Indeed, it might contribute to the `AA` cross-section. The only thing you can really rule out is the `VA` cross-section, because it would have an odd number of $\gamma_5$ in a trace. But if you are not talking about a cross-section, there is no trace yet, so you can not make considerations about $\gamma_5$ being even or odd, since you have to consider the complex conjugate as well.

If this is at linear level, the only thing you can do is to split the vectorial and axial component.

felixhekhorn: @giacomomagni this discussion only concers `fl11`, right?

AleCandido: > @giacomomagni this discussion only concers `fl11`, right?

In that case, @felixhekhorn is right: the axial coupling should not be averaged, since there it would correspond to a loop with a single $\gamma_5$ (just one EW boson per quark-loop).

giacomomagni: > @giacomomagni this discussion only concers `fl11`, right?

This affects all `fl11, flg` and `flps`.

felixhekhorn: i.e. with the paper notation: `fl11` which couples with the "QED singlet" and thus in Vogt language has two components "non-singlet" and "singlet" (meaning pure-singlet) and `fl11g` which is the gluonic counterpart, right?

all in all: in concerns all diagrams with 2 distinct fermion loops, with the boson coupling only to a single loop - now the vanishing of a single $\gamma_5$ terms should apply to each fermion loop separately and so to all 3 channels

giacomomagni: All good :)

giacomomagni:
Are we fine in merging this?

felixhekhorn:
Can you fix pre-commit as much as you can? i.e. whitespace, black, isort

AleCandido:
> Are we fine in merging this?

If you can convince someone to review it would be perfect, otherwise merge it after `pre-commit`.

src/yadism/coefficient_functions/light/n3lo/common.py
```diff
@@ -8,18 +9,61 @@
...
+    if skip_heavylight:
+        w_pc[-1] = 0
+    if not w_pc.any():
+        return 0
+    if channel == "ns":
+        return fl(w_pc)
+    if channel == "s":
+        return fls(w_pc) - fl(w_pc)
+    if channel == "g":
+        return flg(w_pc)
```
felixhekhorn, line 46: Maybe add an exception as last case

giacomomagni: you mean raise an error?

src/yadism/coefficient_functions/light/n3lo/common.py
```diff
@@ -44,6 +44,8 @@ def nc_color_factor(coupling_constants, nf, channel, skip_heavylight):
         return fls(w_pc) - fl(w_pc)
     if channel == "g":
         return flg(w_pc)
+    else:
+        raise ValueError(f"NC color factor {channel} is not defined.")
```
felixhekhorn, line 48:
```suggestion
    raise ValueError(f"NC color factor {channel} is not defined.")
```

felixhekhorn (approved):

PR #184, opened 2023-03-22
Fix sign in rensv

andreab1997:
Following the note wrote by @felixhekhorn (https://github.com/NNPDF/papers/tree/master/nnpdf40mhou), it turns out that there is at least one wrong sign in our implementation of renormalization scale variations. Indeed, if one performs the test of looking at the ratio of NNLO shifts over NLO shifts, one gets a nice decreasing log behaviour with the correct sign and the opposite with the current sign. 
Following the note there is the same problem also at N3LO. @giacomomagni @AleCandido @felixhekhorn

### DIFF
--- a/docs/source/theory/scale-variations.rst
+++ b/docs/source/theory/scale-variations.rst
@@ -29,7 +29,7 @@ Description
 
 Following a more in depth description about the structure of the scale
 variations in the theory.
-This is really decoupled from the process itself, and it has been use as a
+This is really decoupled from the process itself, and it has been used as a
 guidance for the implementation itself, whose details are instead described in
 the :ref:`corresponding technical section <implement-sv>`.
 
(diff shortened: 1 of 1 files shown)


AleCandido:
The unary `+` is more or less an identity function. If you want to keep it, no problem, but you can also consider dropping it :)

giacomomagni:
> Following the note wrote by @felixhekhorn (https://github.com/NNPDF/papers/tree/master/nnpdf40mhou), it turns out that there is at least one wrong sign in our implementation of renormalization scale variations. Indeed, if one performs the test of looking at the ratio of NNLO shifts over NLO shifts, one gets a nice decreasing log behaviour with the correct sign and the opposite with the current sign. Following the note there is the same problem also at N3LO. @giacomomagni @AleCandido @felixhekhorn

Okay by at "least one minus sign" do you mean the note is still WIP so there might be changes??
If this is the case just mark this as draft. 

Would you mind fixing also the [docs](https://yadism.readthedocs.io/en/latest/theory/scale-variations.html), adding an explanation of the origin of this difference especially wrt to the main reference quoted there: https://arxiv.org/abs/hep-ph/0006154

andreab1997:
> > Following the note wrote by @felixhekhorn (https://github.com/NNPDF/papers/tree/master/nnpdf40mhou), it turns out that there is at least one wrong sign in our implementation of renormalization scale variations. Indeed, if one performs the test of looking at the ratio of NNLO shifts over NLO shifts, one gets a nice decreasing log behaviour with the correct sign and the opposite with the current sign. Following the note there is the same problem also at N3LO. @giacomomagni @AleCandido @felixhekhorn
> 
> Okay by at "least one minus sign" do you mean the note is still WIP so there might be changes?? If this is the case just mark this as draft.

No sorry I was unclear. The additional minus signs I was referring to are the N3LO ones that I fixed already. So there should not be other minus signs.
> 
> Would you mind fixing also the [docs](https://yadism.readthedocs.io/en/latest/theory/scale-variations.html), adding an explanation of the origin of this difference especially wrt to the main reference quoted there: https://arxiv.org/abs/hep-ph/0006154

Yes I will do as soon as possible

andreab1997:
> https://arxiv.org/abs/hep-ph/0006154

I did again all the calculations and indeed the new expressions match perfectly with the expressions in https://arxiv.org/abs/hep-ph/0006154 (and so also with the docs) if one correctly take into accounts that:

1. LM and LR become actually -LM and -LR in our notation (as also correctly pointed out by @felixhekhorn ) because in the former case we put the factorization scale at the numerator and in the latter case we put the renormalization scale in the numerator.
2. While the beta functions follow exactly the `EKO` convention (so with the minus sign), when converting the splitting functions P to the anomalous dimensions, if one follows the eko conventions should add a minus sign (of course this is only relevant forf factorization scale variations)

I checked all the expressions up to N3LO and they match perfectly so I am now updating the docs. BTW the docs are not wrong, they are just missing the step splitting functions - > anomalous dimensions and they are using a different definition for the logs

giacomomagni (approved):

andreab1997:
Can we merge this? @AleCandido @felixhekhorn

AleCandido:
> Can we merge this? @AleCandido @felixhekhorn

Up to @felixhekhorn

felixhekhorn (approved):

PR #191, opened 2023-05-17
Polarized matching

giacomomagni:
Here we provide the polarised matching conditions for `g1` . 

@t7phy you can find some `TODO` where you should add the `x-space` expression obtained with Mathematica.
If you can avoid repeating (as I mentioned I think the LL part will be same as the unpolarised one, but plese double check!! )


src/yadism/coefficient_functions/fonll/g1_nc_raw.py
```diff
@@ -388,3 +392,50 @@ def c2g_NNLL_reg(z, _args):
...
+    )
+    return TR * CF * aqq_reg + 1 / 4 * beta0Q * 2 * lo.pqq_reg(z, args) * zeta2
+
+
+@nb.njit("f8(f8,f8[:])", cache=True)
+def pdf_matching_NNLL_sing(z, args):
+    aqq_sing = (
+        -224 / 27 - 4 / 9 * np.pi**2 - 40 * np.log(z) / 9 - 4 * np.log(z) ** 2 / 3
+    ) / (z - 1)
+    return TR * CF * aqq_sing + 1 / 4 * beta0Q * 2 * lo.pqq_sing(z, args) * zeta2
```
giacomomagni, line 441: I think the convention of adding the factor of 2 (both here and in eko) in front of each splitting function coming from https://arxiv.org/pdf/2211.15337.pdf make sense.
As you can see from the tests the LL and NLL pieces are the same as in the unpolarised case and they do match the eko implementation. 
So I'd say from this point of view we are cool with all the conventions!

NOTE: for completeness I believe that in https://arxiv.org/pdf/2211.15337.pdf, splitting functions are following the convention of https://arxiv.org/pdf/hep-ph/9508347.pdf where the DGLAP solution (2.10) contains an addition factor of 1/2 in the exponent.

Radonirinaunimi: Great! So this then solves the mystery we talked about yesterday.

tests/yadism/cf/test_matching_polarized.py
felixhekhorn, line 1: Can we refactor the utility functions (`yad_convolute`, `inverse_mellin`) in `test_matching` so we don't repeat them here?

giacomomagni: Actually I realised we were doing even too much work! 
Due to Ward identities the full `A_{qq,NS}` is the same as in the unpolarised case, so we don't have to add anything. 
For the time being I added a test to cross-check it also with the eko implementation.

src/yadism/coefficient_functions/fonll/g1_nc.py
```diff
@@ -57,38 +57,19 @@ def NNLO(self):
...
-            raw.pdf_matching_NNLL_sing,
-            raw.pdf_matching_NNLL_loc,
-        )
-
+class PdfMatchingNNLLNonSinglet(pc.PdfMatchingNNLLNonSinglet):
+    pass
 
-class PdfMatchingNNNLLNonSinglet(pc.PartonicChannelAsy):
+class PdfMatchingNNNLLNonSinglet(pc.PdfMatchingNNNLLNonSinglet):
     pass
```
felixhekhorn: this might be a left-over from when we had the mismatch of the log order (do you remember @giacomomagni ?) - there should be no N3LL stuff (neither here nor for unpolarized); and indeed they are empty. These would become relevant with FONLL-D/E/F. Please delete both

giacomomagni: wait that was added to make the N3LO working, because of this: https://github.com/NNPDF/yadism/blob/49446b1980b9657131f67d91a4043d993a3b223b/src/yadism/coefficient_functions/fonll/kernels.py#L83-L90

felixhekhorn: Okay, okay, you're right we need it! Exactly as I said "relevant with FONLL-D/E/F" sorry

felixhekhorn (approved):

Radonirinaunimi (approved):

PR #195, opened 2023-05-24
Numerical FONLL

RoyStegeman:
Implements numerical FONLL - supersedes #175


RoyStegeman:
For now I just went with a quick-and-dirty implementation where the output are just the asy terms as implemented, this implementation is mainly for my own understanding. I get some qualitative agreement with apfel though in most of the kinematic domain at least ~10% off. 

Before I proceed though, @AleCandido @felixhekhorn, do you know if the definition of the FFN0 scheme in apfel is the same as what we need? And if so, what is the level of agreement I should find?

src/yadmark/benchmark/runner.py
```diff
@@ -53,7 +53,8 @@ def run_me(self, theory, ocard, pdf):
             method = runcards.Legacy.MOD_EV2METHOD.get(theory["ModEv"], theory["ModEv"])
             method = dictlike.load_enum(types.EvolutionMethod, method)
             method = couplings_mod_ev(method)
-            masses = [mq.value**2 for mq in new_eko_theory.quark_masses]
+            masses = [np.inf, np.inf, np.inf]
```
AleCandido, line 56: ?

If you want to use FFNS3, you should change the `thresholds_ratios`, not the masses.

src/yadism/observable_name.py
AleCandido, line 1: @RoyStegeman @felixhekhorn we should redesign the mechanism for identifying observables, it is definitely exploding...

AleCandido: Even though the `asy` part we can simply lift in the theory, to the FNS.

RoyStegeman: I agree that the mechanism for identifying observables as it is, is not ideal. Not sure what you mean by lifting it to FNS? The problem with that is that based only on the active flavors (as FNSs are implemented now) one cannot distinquish FFN0 from FFNS so that's why I went implemented it as an independent observable (though agree it's not super clean).

AleCandido: I'm proposing to use not the `NfFF` key (the active flavors, that we actually implement with thresholds) but the `FNS` key, in the very same way of APFEL:

```
FNS: FFNS
NfFF: 4
``
is FFNS4, while:
```
FNS: FFN0
NfFF: 4
```
is FFN04.

Comment (RoyStegeman) [reply src/yadism/observable_name.py:1]: > I'm proposing to use not the `NfFF` key

This seems opposite of what you propose, do you mean that we should use the `NfFF` key instead of **not** using it? But that's what we do now if I understand you correctly. Much of the mess comes from the fact that FFN0x and FFNSx have the same thresholds so we need some other way to distinguish them.

Comment (AleCandido) [reply src/yadism/observable_name.py:1]: I'm really proposing to use the `FNS` key, **not** the `NfFF`. I was using the latter just to avoid writing the thresholds. Let me write it explicitly.

```
FNS: FFNS
kDIScThr: 0
kDISbThr: inf
kDIStThr: inf
```
is FFNS4, while:
```
FNS: FFN0
kDIScThr: 0
kDISbThr: inf
kDIStThr: inf
```
is FFN04.

The keys discussed are available in https://docs.nnpdf.science/theory/theoryindex.html. `kDISxThr` is an extension, but unneeded at DB level, since we could determine from the keys present (in particular, `FNS: FONLL` would be compiled into the others, with suitable matching scales).

Soon I'll also rework the `input` layer (i.e. dropping it, it's heavy and not very useful, instead I will use some runcards objects, as for EKO), I'll try to simplify this kind of things.

Comment (RoyStegeman) [reply src/yadism/observable_name.py:1]: But this is still what is done currently, except at the level of the runcard we set `NfFF` which in the first few lines of the yadism runner is used to set the thresholds you propose. So the difference is that currently `kDIS{fl}Thr` are hidden from the user. 

The problem I see is that currently `Combiner` has knowledge of the observable and the masses but not the `FNS`. Actually checking it now I see it should have the `FNS` through `ESFInfo`, so we can read that in `Combiner` as well, which is cleaner than appending 'asy'. 

So in that case I would propose to drop the asy observable name (and all support for it) and call the observable `F_charm` regardless of FFN0 or any other scheme (which is consistent since we also don't give a different name for FONLL or FFNS).

Comment (RoyStegeman) [reply src/yadism/observable_name.py:1]: I think the first paragraph is also what you meant, but then I still don't understand what you want to change with respect to how it's done now (unless you meant the same as my second paragraph).

Comment (AleCandido) [reply src/yadism/observable_name.py:1]: The proposal should be exactly to use the `FNS` instead of `obsname.flavor`.

So, the observable name will be reserved for true kind (`F2`/`FL`/`XSHERANC`/...) and true flavor (`charm`/`bottom`/`top`/`total`).

The only thing I'm truely undecided upon is whether we should keep or not the `light` flavor, and with which definition (`uds` or zero-mass).

Comment (AleCandido) [reply src/yadism/observable_name.py:1]: > But this is still what is done currently, except at the level of the runcard we set `NfFF` which in the first few lines of the yadism runner is used to set the thresholds you propose. So the difference is that currently `kDIS{fl}Thr` are hidden from the user.

This is just legacy mode:
https://github.com/NNPDF/yadism/blob/master/src/yadism/input/compatibility.py#L6
but it was always intended as a device for a transition, not to stay there forever.

What I intended was exactly your second and third paragraph.

Comment (RoyStegeman) [reply src/yadism/observable_name.py:1]: > but it was always intended as a device for a transition, not to stay there forever.

I see, so scrap `NfFF` in favor of the thresholds. Not sure if I would include that in this PR, but since yadism consumes the thresholds I'd say it makes more sense to do it that way. The translation would only really be needed for benchmarking exercises.

Comment (RoyStegeman) [reply src/yadism/observable_name.py:1]: Okay good, so we agree on what to do for  the observable name. I'll change it tomorrow.

> The only thing I'm truely undecided upon is whether we should keep or not the `light` flavor, and with which definition (`uds` or zero-mass).

I'm not sure I understand what you mean by this. Are you considering scrapping `ligh` for only `total`?

Comment (AleCandido) [reply src/yadism/observable_name.py:1]: Yes, `light` has been controversial since long.

APFEL considered as `light` the contributions in which one of `{up, down, strange}` is coupling to the EW boson. That is a physically viable definition, but contains mass effects.
Instead, to have a pure ZM observable, we defined `light` flavor exactly as the observable without mass effects, i.e. where masses of active flavors are all 0.

However, I'm no longer sure it has been a sensible choice: this can also be achieved selecting the FNS (and, in practice, tuning the masses values), and APFEL definition of `light` is the only sensible one for FONLL, that lead us to have two definitions of `light` in yadism:
- ZM observable in general
- `up + down + strange` in FONLL, a la APFEL

I'm thinking if we should actually allow for the APFEL definition, or just no definiton at all, and keep `light` as internal (exposing only `total`, as you said).

Comment (RoyStegeman) [reply src/yadism/observable_name.py:1]: I'm feel a bit out of my depth with that decision, so I'm happy with whatever you decide (though if we change the definition we should have a good reason for it). What is the use-case for exposing `light` in either definition? If there is none, or just because APFEL does it and historically you decided to follow that, it may indeed be best to only expose `total`.

**Review by AleCandido — COMMENTED** [review]
Most of the files have small modifications, and the correct ones.

Now I'm on the phone, so I'm not really comfortable with the large files. But I'll try to finalize later today the review for the combiner and kernels. Apparently they look good as well.

File: src/yadism/input/compatibility.py [hunk 2]
```diff
@@ -131,9 +129,4 @@ def update_fns(theory):
     for fl in hqfl:
         theory[f"kDIS{fl}Thr"] = theory[f"k{fl}Thr"]
 
-    # fix FONLL-B and introduce PTODIS
-    if fns == "FONLL-B":
-        theory["PTODIS"] = 2
-        theory["PTO"] = 1
-    else:
-        theory["PTODIS"] = theory["PTO"]
+    theory["PTODIS"] = theory["PTO"]
```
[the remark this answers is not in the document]
AleCandido: Great, thanks :D

src/yadism/observable_name.py
felixhekhorn: #198

src/yadism/coefficient_functions/__init__.py
```diff
@@ -67,34 +65,23 @@ def light_component(self):
...
-            comp.extend(
-                fonll.kernels.generate_light(
-                    self.esf,
-                    nl,
-                    self.esf.info.theory["pto_evol"],
-                    self.esf.info.theory["pto"],
-                )
-            )
-            comp.extend(heavy.kernels.generate_missing(self.esf, nl, nl + 1))
+        if self.scheme == "FFN0":
```
felixhekhorn, line 68: This does not look correct: light does not depend on any masses and so not on any logs and so is just the same in `FFNS` or `FFN0`

felixhekhorn: The exception is of course missing - but the branching is only needed for those and nowhere else

felixhekhorn: In practice this means `comp.extend(light.kernels.generate(self.esf, nf))` (line 69) should move before the branching and `asy.kernels.generate_light_asy` needs to be renamed to `generate_missing_asy` and rewritten (i.e. it will become much easier

RoyStegeman: The `generate_light_asy` function is just `light.kernels.generate`, `missing`, `damping`, and an additional term at N3LO since $F_d$ doesn't vanish from that order. The  `light.kernels.generate` could indeed be moved out of the asy function, but missing wouldn't be quite the correct name since there is still the N3LO contribution

Actually, this makes me think - we should now also allow for the damping to be applied for FFNS since we may want to that to generate a contribution to $F_d$ in FONLL.

felixhekhorn: I commited something in https://github.com/NNPDF/yadism/pull/195/commits/b828f1222f09781b82bb4d5b8b1a7a48269083c9 (untested) ... I believe this should go into the correct direction ...

RoyStegeman: Thanks! Shouldn't the order as^3 term stay at least somewhere though? And do you agree regarding the damping in FFNS?

src/yadism/observable_name.py
```diff
@@ -120,18 +105,13 @@ def is_heavy(self):
             is_heavy : bool
                 is a heavy flavor?
         """
-        return self.flavor != "light"
+        return "light" not in self.flavor
```
AleCandido, line 108: Are we sure?

The difference from the previous one and this one is that `heavylight` was previously *heavy*, and now it is not any longer.

In principle, we might not need `heavylight` any longer. In which case it's just the same (but, if we're still using it, we should make sure of the effect of this).

src/yadism/observable_name.py
```diff
@@ -140,15 +120,13 @@ def is_heavylight(self):
     @property
     def is_composed(self):
         """Is it a composed flavor? i.e. total"""
-        return self.flavor == "total"
+        return "total" in self.flavor
```
AleCandido, line 123: This one should be the exact same, since only `"total"` contains `"total"`. And the former one is more restrictive.

I would keep the old one.

src/yadism/coefficient_functions/__init__.py
```diff
@@ -43,58 +43,47 @@ def __init__(self, esf):
...
-
         family = self.obs_name.flavor_family
-
         # Adding light component
         if family in ["light", "total"]:
             comps.append(self.light_component())
-        if family == "heavy":
-            #  the only case in which an heavy contribution is not present in those
-            #  accounted for in total, it's whene heavy already became heavylight
+        if family == "heavy" and self.scheme != "FFN0":
```
AleCandido, line 54: Most likely, we should discuss this part in more details.

src/yadism/coefficient_functions/__init__.py
```diff
@@ -119,56 +108,60 @@ def heavy_components(self):
...
+        # we are above the mass threshold. For these quarks the masses are not
+        # considered.
         for sfh in range(nf, 7):
             # exclude sfh=3, since heavy contributions are there for [4,5,6]
             # if it's ZM you don't even have the component
-            if sfh not in masses or not masses[sfh]:
+            if sfh not in masses:
+                continue
+
+            # There is no massive heavy contribution for ZM
```
AleCandido, line 120: Thanks for making all of this more understandable :D

src/yadism/coefficient_functions/__init__.py
felixhekhorn: - concerning damping: asy-s need to be damped for sure, then for the rest I'll write a long comment in a second on the PR
- concerning N3LO: mmm @giacomomagni is the best to comment, but I think in Numerical FONLL there is no need for a special treatment any longer ... @giacomomagni can you remind me what the problem is?

AleCandido: > concerning N3LO: mmm @giacomomagni is the best to comment, but I think in Numerical FONLL there is no need for a special treatment any longer ... @giacomomagni can you remind me what the problem is?

I don't know about N3LO specifically, but I agree there should not be any special treatment any longer...

felixhekhorn:
There is a complication which I believe we don't take care of:

Consider the measurement of three ESF: a) Q2 = 1 GeV² b) Q2 = 10 GeV² c) Q2 = 100 GeV² - how to compute the theory prediction for these three points with different settings? and by that I mean global theory settings (so before any theory remapping)

1. ZM-VFNS:
  a) $c_l(3)*f(3)$
  b) $c_l(4)*f(4)$
  c) $c_l(5)*f(5)$ with $c_l$ = purely light = Vogt
2. FFNS3 a.k.a. ABM
  a) $c(3)*f(3)$
  b) $c(3)*f(3)$
  c) $c(3)*f(3)$ with $c = c_l + c_h$ and $c_h$ = my thesis
3. FFNS4
  a) $c(4)*f(4)$
  b) $c(4)*f(4)$
  c) $c(4)*f(4)$
4. FONLL
  a) $c(3)*f(3)$ in three flavors we have to fall back to FFNS3!
  b) $c(3)*f(3) + D[c(4)*f(4) - c^a(3)*f(3),3]$ where $D$ refers to damping and $c^a = c_l^a + c_h^a = c_l + c_h^a$ is asy; Note that $c(4)$ for ESF a) was empty!
  c) $c(3)*f(3) + D[c(4)*f(4) - c^a(3)*f(3),3]  + D[c(5)*f(5) - c^a(4)*f(4),4]$ again, $c(5)$ has been empty for ESF a) and b); I'm not entirely sure on that formula either, since I somehow assume that the Ds commute,which might or might not be true ... recall that the default matching depends on a quark mass (hence the parameter $D[...,n_f]$

The problem with the current implementation (and general scheme) is that we can not distinguish scenario 3 and 4 for ESF a) - we need a flag in the theory run card to do this ... in the "FONLL-FFNS4" runcard we need to silence $c(4)$ below $m_c$

Likewise we need a way to not damp in "FONLL-FFNS3", but in all others (i.e. "FONLL-FFNS4" etc.)

AleCandido:
> The problem with the current implementation (and general scheme) is that we can not distinguish scenario 3 and 4 for ESF a) - we need a flag in the theory run card to do this ... in the "FONLL-FFNS4" runcard we need to silence

This I do not understand: we do not have to distinguish anything, yadism is only providing the ingredients for the combination, and they do not need to know that will be used for FONLL later on, at all.

In case of FFNS4 a), just works. In case of FONLL a), also works, because that's the FONLL prescription.
If you want the "sliding" FONLL it should be implemented at combination time, just setting to 0 a part of the grid.

src/yadism/coefficient_functions/__init__.py
giacomomagni: I'm not sure I follow entirely the discussion. If I assume $F_d = F^{(nf+1)} -F^{(nf),0} $, 
it is not vanishing already at NNLO (due $C_{q,{NS}}$). 
The only difference at N3LO is that now also $C_g$  and $C_{q,PS}$ should contribute to  $F_d$. 
However we know these coefficients only for the massive $F_{heavy}$ (and only through the approximation of Nicolò), while massive corrections to $C_{i,light}, \ i= {NS,g,PS}$ are unknown.

RoyStegeman: We specifically mean the light coefficient contribution below:
https://github.com/NNPDF/yadism/blob/1bf52faf2eccde3a3049ebd63b94266148c1035d/src/yadism/coefficient_functions/fonll/kernels.py#L50-L58

I don't know where this term originates and if it should be kept (as a contribution to asy?) for numerical fonll.

RoyStegeman:
> If you want the "sliding" FONLL it should be implemented at combination time, just setting to 0 a part of the grid.

Do you mean putting a step function at the thresholds? That should indeed be possible at combination time, but I don't think this is the case for the non-heaviside component of the damping factor i.e. the $(1-\frac{Q^2}{m^2})^2$.

AleCandido:
I believe that @felixhekhorn is really talking about the Heaviside part, i.e. "below threshold" (below the matching scale).

He was arguing that below we should only have a single grid, the FFNS3, while above the matching we need the three of them, FFNS3 - FFNS30 + FFNS4.
To me, this is an unrequired complication. I would simply go for 3 grids all the time, paying the price to compute the extra part (DIS is cheap anyhow, and light DIS is particularly inexpensive to compute, and these points are usually a handful). The combiner should know what to set to 0, since this is part of the FONLL prescription.

src/yadism/coefficient_functions/__init__.py
giacomomagni: I see, this fix was needed because following the [FONLL paper](https://arxiv.org/pdf/1001.2312.pdf), the cancellation:

https://github.com/NNPDF/yadism/blob/1bf52faf2eccde3a3049ebd63b94266148c1035d/src/yadism/coefficient_functions/fonll/kernels.py#L46-L48

has been implemented. Up to NNLO, it turns out that the light elements can be generated with `nl` instead of `nl+1`.  

This is not true anymore for the N3LO contributions, because the "derivative argument" of the paper (see section 3.2.2 ) do not hold, so one need to properly add the ZM contribution using $F^{(n_l+1)}$. 
This makes a difference because at N3LO the coefficients have an explicit dependence on `nf` which do not factors out anymore (for pure singlet and gluon).

While adding this contribution we skip the `heavy_light` because that piece is already taken care of by: 

https://github.com/NNPDF/yadism/blob/1bf52faf2eccde3a3049ebd63b94266148c1035d/src/yadism/coefficient_functions/fonll/kernels.py#L163

If I understand correctly, in numerical FONLL, the light contribution will be always generated with `nl+1` active flavors, dropping the cancellation done in  FONLL paper;  so this trick should not be needed. 
Hope this can clarify the situation :)

RoyStegeman:
Ah I hadn't fully understood that was what @felixhekhorn is suggesting, but since it's not that expensive I agree that we may as well do the damping (heaviside only) outside of yadism and just set the components of the fktables contributing to diff below the relevant threshold to 0. Probably that's more convenient than requesting only the specific Q2 values from yadism. 

What should we then do with the damping factor? Apply the damping factor at all scales and then set to 0 the contributions below thershold or simply keep only the heaviside?

src/yadism/coefficient_functions/__init__.py
RoyStegeman: Thanks for the clarification, in that case it can indeed be dropped.

src/yadism/coefficient_functions/light/kernels.py
```diff
@@ -60,7 +60,7 @@ def generate(esf, nf, skip_heavylight=False):
         skip_heavylight=skip_heavylight,
     )
 
-    # remove also singlet contributions which is added by generate_light_diff
+    # remove also singlet contributions which is added by generate_light_asy
```
giacomomagni, line 63:
```suggestion
    # remove also singlet contributions which is added while subtracting light_asy with light
```

src/yadism/coefficient_functions/__init__.py
```diff
@@ -120,56 +109,60 @@ def heavy_components(self):
...
+            if not masses[sfh]:
                 continue
 
             heavy_comps[sfh] = Component(sfh)
+
+            # calculate only the contribution corresponding tho the observable
```
giacomomagni, line 127:
```suggestion
            # calculate only the contribution corresponding to the observable
```

AleCandido:
> What should we then do with the damping factor? Apply the damping factor at all scales and then set to 0 the contributions below thershold or simply keep only the heaviside?

I'm now wondering: can't we apply the whole damping in the combiner?

tests/yadism/cf/test_nc_asy_polarized.py
```diff
@@ -1,130 +1,129 @@
 import numpy as np
 from utils import MockESF
 
-from yadism.coefficient_functions.fonll import g1_nc as f_g1_nc
-from yadism.coefficient_functions.fonll import g1_nc_raw
+from yadism.coefficient_functions.asy import g1_nc as f_g1_nc
+from yadism.coefficient_functions.asy import g1_nc_raw
 from yadism.coefficient_functions.heavy import g1_nc as h_g1_nc
 
+# class Test_g1_raw:
```
giacomomagni, line 8: This test should not be removed, it tests the `g1` asymptotic.

RoyStegeman: That's why I uncommitted it instead of deleting it, I just didn't want the spam from the failing tests

RoyStegeman:
So isn't it an option to calculate the three fktables for all $Q^2$ but before summing them together using pineko we multiply by 0 all bins below the threshold in two of the fktables? That way we wouldn't need damping in yadism and we avoid the problem with inconsistent grid sizes. It would mean that damping is heaviside-only, which I don't know how much of a problem that would be.

Are the Q2 of the datapoints stored in the fktables?

AleCandido:
> It would mean that damping is heaviside-only, which I don't know how much of a problem that would be.

Can't we apply non-Heaviside damping as well? In the end, damping is just a pre-factor, and depends on variable that are still available in the resulting grid, isn't it?

RoyStegeman:
According to @felixhekhorn not, though I'm not sure I still agree/understand his argument since the damping factor only depends on Q2 and the mass. Perhaps he can clarify his argument here.

felixhekhorn:
> According to @felixhekhorn not, though I'm not sure I still agree/understand his argument since the damping factor only depends on Q2 and the mass. Perhaps he can clarify his argument here.

We can delegate damping to pineko if the damping does only act on $Q^2$ - this is not necessarily the case: see Eq. (20) of the FONLL paper - you can also do a rescaling of $x\to\chi$ (and by chance JR was asking for just that on Friday for the polarized paper); for such a scenario you would also need to alter the obs cards ... (more generally speaking the only condition for damping is Eq. (18), so one could even cook up a partonic prescription) ... However, I think in practice we never do such a thing (and I believe APFEL can only do the Heavyside function)

AleCandido:
$\mu^2$ and $z$ are both available in the subgrids, so this is perfectly available (just need to use the new PineAPPL tools to edit subgrids).

In principle, even Bjorken-$x$ is available, but this quantity is DIS-specific, while FONLL is not a DIS-specific prescription, so I do not expect to be relevant. Indeed, it would be available in a non-standard way, as one of the bins-defining observables.
Same story for the actual photon virtuality $Q^2$.

As well known, in most of DIS $x = z$ (at LO) and $Q^2 = \mu^2$ (w/o scale variations), so the content of the bins is also fixing the content of the subgrid. But this is a chance, and we can always rely on the subgrid structure.

RoyStegeman:
Thanks, in that case I'll just drop damping here and do it in pineko.

AleCandido:
@RoyStegeman is there a specific reason why this is still *draft*? To me it looks close to completion

RoyStegeman:
I was actually just thinking the same :).  Currently double checking if I am happy about the intrinsic bit since now I only removed the minus signs without really understanding which terms are there, but maybe you can confirm that that's what should be done?

AleCandido:
> Currently double checking if I am happy about the intrinsic bit since now I only removed the minus signs without really understanding which terms are there, but maybe you can confirm that that's what should be done?

I believe the *name* to be still wrong, since it is still set to `Matching*`. But possibly even the implementation.

To me, they seem to be all power-suppressed terms, e.g. a simple one:
https://github.com/NNPDF/yadism/blob/798882ab323c5dbc0a8f254283edc87b87e6fba9/src/yadism/coefficient_functions/intrinsic/raw_nc.py#L90-L91
And, even though I can't motivate why there is no $log(Q^2)$, if they are absent they have no impact on the FONLL prescription. I.e. the intrinsic contribution will only be active in FFNS3, and zero for FFNS30.
(I believe that might be related to factorization, but still the initial mass should protect collinear logs...)

Maybe @felixhekhorn can confirm this by heart, but I might be missing some insight...

RoyStegeman:
@felixhekhorn having a quick look at the Ball, Bonvini, Rottoli paper I think the asy components for intrinsic or not are the same in both cases and that the difference between the two schemes is an additional heavy contribution (`intrinsic.kernels.generate`) and a different matching as done in `generate_heavy_intrinsic_diff`. I think the matching can be removed so the asy component for intrinsic is the same as extrinsic. Can you confirm?

I could spend today going through the paper but someone here must know the answer right?

AleCandido:
> I think the asy components for intrinsic or not are the same in both cases

I don't believe so, because intrinsic are rather different diagrams (massive particle incoming), so, if they have an asymptotic limit, it will be different from heavy production. Or at least I don't see a reason for them to be the same. Am I missing something? @felixhekhorn

About this, BBR has no discussion about the structure of the contribution, nor any explicit coeff func.


> I think the matching can be removed so the asy component for intrinsic is the same as extrinsic. Can you confirm?

About this I can fully confirm. The goal of *numerical* FONLL is really to abolish the matching in coeff funcs, and keep it on the evolution side.
The whole *numerical* FONLL project started from an observation in BBR:

> One way of using the FONLL construction would be simply to compute the three ingredients
F (3), F (3,0) and F (4), using the factorized expression in the 3-flavour and 4-flavour schemes,
Eqs. (2.1), (3.2), (2.8), and combine them linearly according to Eq. (3.3). In practice this
is awkward, because it means one has to work simultaneously with PDFs in two different
schemes. 
> <p align="right"><a href="https://arxiv.org/abs/1510.02491">Ball, Bonvini, Rottoli</a>, Section 3.2</p>

Where they use this observation (essentially "numerical FONLL is awkward") to motivate the rest of the paper. So, for the implementation of numerical FONLL, the paper content is not terribly useful.

felixhekhorn:
Oookay - so

> I believe the name to be still wrong, since it is still set to Matching*.

yes, the name is weird - however, I guess the name `PartonicChannelAsyIntrinsic` was already taken so we needed something else :upside_down_face: - see also my comment below

> I could spend today going through the paper but someone here must know the answer right?

if it were not so long time ago I would know :innocent: - let me try to guess from what is implemented instead and reading the relevant papers diagonally ...

> And, even though I can't motivate why there is no log(Q2)

there is - here: https://github.com/NNPDF/yadism/blob/d620015be9c3dfa9c2f8beeb3194889843a4a892/src/yadism/coefficient_functions/fonll/partonic_channel.py#L222
these coefficient functions depend on $Q^2/m^2$ and so they survive the asymtotic limit and so they gain a log ...
I think the relevant equation is eq. (12) in [Kretzer&Schienbein](https://arxiv.org/abs/hep-ph/9805233) (our bad to not to spell this in the docs there)

> I think the matching can be removed so the asy component for intrinsic is the same as extrinsic.

with what I just said: no. Moreover I can see in the code https://github.com/NNPDF/yadism/blob/d620015be9c3dfa9c2f8beeb3194889843a4a892/src/yadism/coefficient_functions/fonll/partonic_channel.py#L227
a dependency on the parent local part (which is surely individual)

I also think that the gluonic part:
https://github.com/NNPDF/yadism/blob/d620015be9c3dfa9c2f8beeb3194889843a4a892/src/yadism/coefficient_functions/fonll/partonic_channel.py#L283
can be dropped - I believe this entered via the scheme change, aka matching (most likely why we named it "Matching") ... however the quark part should have a genuine asy part

src/yadism/coefficient_functions/asy/partonic_channel.py
felixhekhorn, line 1: `Delta_qq_sing`, `Delta_qq_loc`, `K_qq_sing` and `K_qq_loc` can be dropped - since the first to are the contribution to the scheme change by $P_{qq}$ (see eq. (38) of the FONLL paper) and the last two the explicit matching conditions (eq. (33) ibid.)

src/yadism/coefficient_functions/__init__.py
felixhekhorn: to be more specific: when we need to compute F2c in FFNS4 and FFN04 they are just the same: heavylight

RoyStegeman: Ah yes you may be right. The reason I left it out was that in the current master implementation it is only evaluated once in the FONLL scheme so I figured any cancellation of this term would be in the `*diff` function and thus including this in FFN0 would result in a double counting. However, this term should probably cancel between the contributions in three flavor scheme (in which I don't believe it contributes, but still the exception for FFN0 shouldn't be here) and then result in a net contribution for the 4 flavor scheme.

src/yadism/coefficient_functions/__init__.py
```diff
@@ -44,58 +44,44 @@ def __init__(self, esf):
...
+                            ihq,
+                            self.esf.info.theory["pto_evol"],
+                        )
+                    )
+                else:
+                    comp.extend(heavy.kernels.generate_missing(self.esf, nf, ihq))
+
         return comp
 
     def heavylight_components(self):
```
felixhekhorn, line 87: to open another non-trivial point: since the whole point of NumFONLL is to account for simultaneous mass-effects we could/should make also heavylight more complicated and account for "higher"-mass effects: i.e. bottom-mass effects to F2c in FFNS4. This can be done using the `icoupl` argument of [`generate_missing`](https://github.com/NNPDF/yadism/blob/bc4ec4c9730724be453fea3e63d751b32cf3223a/src/yadism/coefficient_functions/heavy/kernels.py#L79) (and similar for asy)

AleCandido: The goal of numerical FONLL is to simplify, not to complicate, definitely.

However, if we're not supporting it yet (it's not one of the features coming out of the box), open a dedicated issue. I would not include more features in this PR.

AleCandido: Thanks!

RoyStegeman:
So the `MatchingGluon` should be dropped while `MatchingIntrinsic` should stay?

felixhekhorn:
> So the `MatchingGluon` should be dropped while `MatchingIntrinsic` should stay?

I think, yes - @AleCandido do you agree? 

On second thought: I also think, that `FMatchingQuark` should change: to the present expression $P_{qq} * \ln(Q^2/m^2)$ should be subtracted - this bit shoud come from the scheme change (same as is for the gluon channel with $P_{qg}$) ... so what I said earlier (that it has a log) is true AnaFONLL, but should not be there in NumFONLL ... recall that in the scheme trafo you have one running eko (generating $P * L$) and one intrinsic eko (being 1)

RoyStegeman:
Please have a look at the last commit to see if that's what you have in mind. I didn't remove the dead code to make it easier to understand the changes, but will do so if this is correct.

src/yadism/coefficient_functions/asy/partonic_channel.py
```diff
@@ -63,7 +61,7 @@ def sing(z, _args):
                 asnorm
                 * icl
                 * constants.CF
-                * ((1.0 + z**2) / (1.0 - z) * (l - 2.0 * np.log(1.0 - z) - 1.0))
+                * ((1.0 + z**2) / (1.0 - z) * (-2.0 * np.log(1.0 - z) - 1.0))
```
AleCandido, line 64: No wait, I'm not sure: if there's a log, the log should stay, isn't it?

Can we have the original expression? It might be in some notebook in this repo, but maybe @felixhekhorn recalls better than me...

felixhekhorn: Now I'm no longer sure any longer ... and in fact it is reintroduced in https://github.com/NNPDF/yadism/pull/195/commits/961357a8940a9af8ff05f34d86b830bb2bb79973 - however, I don't understand why ... I did it because of the game I was playing in [test_nc_asy.py](https://github.com/NNPDF/yadism/pull/195/commits/961357a8940a9af8ff05f34d86b830bb2bb79973#diff-20039f8f2fd50247add715384e577325eaf35a0cc88757e8b51edf6800b05086) in order to prove my previous statement - well it turned out with the log the difference is much better, so maybe I was just wrong and so I reintroduced it ...

On pure theory grounds I stand with what I said [earlier](https://github.com/NNPDF/yadism/pull/195#issuecomment-1580729465) ...

felixhekhorn:
in https://github.com/NNPDF/yadism/pull/195/commits/961357a8940a9af8ff05f34d86b830bb2bb79973 I hopefully fixed the LO behaviour given by Eq. (7) of Kretzer&Schienbein

felixhekhorn:
> in [961357a](https://github.com/NNPDF/yadism/commit/961357a8940a9af8ff05f34d86b830bb2bb79973) I hopefully fixed the LO behaviour given by Eq. (7) of Kretzer&Schienbein

Okay ... that was naive - the equation is the correct one, but then there is a non-trivial mapping between their notation and ours, I forgot ... https://github.com/NNPDF/yadism/pull/195/commits/0f82a19eb8fd076f3ac93f07060e35f38ae216cd should do the correct thing - and it even makes sense

<details>
<summary>MMA snippet</summary>

```Wolfram
Module[{Spp, Smp, Spm, \[CapitalDelta], \[Lambda], y, \[Eta], convp},
 Spp = Qsq + m2sq + m1sq;
 Smp = Qsq - m2sq + m1sq;
 Spm = Qsq + m2sq - m1sq;
 \[Lambda][a_, b_, c_] := a^2 + b^2 + c^2 - 2 (a b + b c + c a);
 \[CapitalDelta] = Sqrt[\[Lambda][m1sq, m2sq, -Qsq]];
 \[Eta] = 2*Qsq/(Spp + \[CapitalDelta]);
 y = -Qsq/m1sq;
 convp = x/\[Eta];
 Print["CC F2 S+ ", 
  FullSimplify[
   Series[(1 - 1/y), {m1sq, 0, 0}, 
    Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}], 
   Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}]];
 Print["NC F2 S+ ", 
  FullSimplify[
   Series[(\[CapitalDelta]/Qsq * \[Eta]), {m1sq, 0, 0}, 
    Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}], 
   Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}]];
 Print["CC F3 R+ ", 
  FullSimplify[
   Series[1, {m1sq, 0, 0}, 
    Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}], 
   Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}]];
 Print["NC F3 R+ ", 
  FullSimplify[
   Series[(x/convp), {m1sq, 0, 0}, 
    Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}], 
   Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}]];
 Print["CC FL S+ ", 
  FullSimplify[
   Series[-1/y, {m1sq, 0, 0}, 
    Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}], 
   Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}]];
 Print["NC FL S+ ", 
  FullSimplify[
   Series[((\[CapitalDelta]/Qsq - Spp/\[CapitalDelta])*\[Eta]), {m1sq,
      0, 0}, Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 
      1 > \[Xi] > 0}], 
   Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}]];
 Print["NC FL S- ", 
  FullSimplify[
   Series[(2*Sqrt[m1sq*m2sq]/\[CapitalDelta]*\[Eta]), {m1sq, 0, 0}, 
    Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}], 
   Assumptions -> {Qsq > 0, m1sq > 0, m2sq > 0, 1 > \[Xi] > 0}]];
 ]
```

</details>

 (+ recall that for NC $m_1=m_2$)

RoyStegeman:
Thank @felixhekhorn! If this solves the intrinsic issue I think all this PR needs is to clean up some of the unused code (and maybe fix some of the docstrings)?

src/yadism/coefficient_functions/asy/partonic_channel.py
felixhekhorn: > Can we have the original expression? It might be in some notebook in this repo, but maybe @felixhekhorn recalls better than me...

I believe we had never the full expression in MMa, but only parts (like fhat), or at least I cannot find it ... I believe we only went through Fortran (and translated from there)

AleCandido:
> maybe we should rename `pto_evol` to something more meaningful ... we still need the information of course, we could call it `the_other_pto`, since to cancel we need to know both PTOs

Do we really need that information?

I believe we don't. If we compute massive, we do it consistently with some flavors. If we compute asymptotic, the same. If we compute zero-mass, again, you only have one `pto`.

When would we need the two different orders now?

felixhekhorn:
> When would we need the two different orders now?

We need them - the case is of course FONLL-B! The FONLL prescription is "FNS3 + FNS4 - DoubleTerms(FFNS3,FFNS4)" - so asy depends on both ... however, thinking one second more ... arrrrrgh - the thing is again more complicated - i.e. FONLL-B is more complicated in NumFONLL ... 

Coming back to my [three ESF from above](https://github.com/NNPDF/yadism/pull/195#issuecomment-1576996347):
1. In FONLL-B we want "light"=the higher FNS to be at NLO (=NLO-DIS) and the "heavy"=the lower FNS to be at at NNLO(=NNLO-DIS=NLO-heavy).
2. if we only think about EFS b): we need FFNS3 with PTO=2 and FFNS4 with PTO=1 - and in FFN03 you only cancel the terms that appear in *explicitly* in FFNS4, which since this is PTO=1 are exactly the NLL terms - so FFN03 has to be PTO=2 and PTO_EVOL=1, letting the [NNLL terms inside FFNS3 survive](https://github.com/NNPDF/yadism/blob/981d8f9c03678292d76e081a754e90bda6ceb76f/src/yadism/coefficient_functions/asy/f2_nc.py#L54)
3.  now, the complication: the decision (which one is NLO and which NNLO) is the same way dynamic the way shutting done some FFNS4 coefficient function is - so the combiner (aka pineko) needs both grids FFNS4(PTO=1) for ESF b) and FFNS4(PTO=2) for ESF c)

PS: actually `pto_evol` is not such a bad name it is reffering to *the evolution log*

AleCandido:
> We need them - the case is of course FONLL-B! The FONLL prescription is "FNS3 + FNS4 - DoubleTerms(FFNS3,FFNS4)" - so asy depends on both ... however, thinking one second more ... arrrrrgh - the thing is again more complicated - i.e. FONLL-B is more complicated in NumFONLL ...

Nono, in "numerical FONLL" (observable-level FONLL), *asy* is incredibly simple: it is just FFNS3 in the $Q^2 \to \infty$ limit. And thus the PTO is only that one of FFNS3, nothing else.

We explicitly discussed with Stefano: FONLL is a very simple idea
- take the full zero-mass (resummed) observable, as is
- add on top the power terms from the massive (FO) one
  - to extract this, you just have to take the massive observable *as is*, and subtract the $Q^2 \to \infty$ limit, the drops everything that is power-suppressed, and not anything else

So, you're just confusing yourself once more: there is no prescription at coefficient functions level. There are just observables, and nothing else.

AleCandido:
> PS: actually `pto_evol` is not such a bad name it is reffering to _the evolution log_

That's why I chose it in the first place. But now it's obsolete, since yadism doesn't care any longer.

RoyStegeman:
I don't really understand what diagrams the `missing` contribution corresponds to. The docs say that they contribute to `F_total` but not light and heavy, though it seems that it always contributes. 

In [20daf43](https://github.com/NNPDF/yadism/pull/195/commits/20daf430a35aee2e1ab58e0658f6eac9e47e5d90) `missing_asy` is added to heavy but this introduces a disagreement with apfel. It makes sense that if there is a missing contribution in FFNS there should be a missing_asy contribution for FFN0, but I would like to understand why apfel disagrees.

AleCandido:
> I don't know what diagrams the `missing` contribution corresponds to. The docs say that they contribute to `F_total` but not light and heavy, though it seems that it always contributes.

That's the exact definition of missing (if you pick a suitable definition of *light*, i.e. zero-mass): they are diagrams in which a massive quark line is present, but it is not coupling to the EW boson.

> In [20daf43](https://github.com/NNPDF/yadism/pull/195/commits/20daf430a35aee2e1ab58e0658f6eac9e47e5d90) `missing_asy` is added to heavy but this introduces a disagreement with apfel. It makes sense that if there is a missing contribution in FFNS there should be a missing_asy contribution for FFN0, but I would like to understand why apfel disagrees.

Actually APFEL should have those contributions as well, even though they might be slightly different in the case of NC, since they have been updated by @felixhekhorn (they should be provided by LeProHQ).
At this point, it seems faster to wait directly the most involved guy to reply ^^

felixhekhorn:
> I don't know what diagrams the `missing` contribution corresponds to. The docs say that they contribute to `F_total` but not light and heavy, though it seems that it always contributes.
> 
> In [20daf43](https://github.com/NNPDF/yadism/pull/195/commits/20daf430a35aee2e1ab58e0658f6eac9e47e5d90) `missing_asy` is added to heavy but this introduces a disagreement with apfel. It makes sense that if there is a missing contribution in FFNS there should be a missing_asy contribution for FFN0, but I would like to understand why apfel disagrees.

The missing diagrams are exactly as @AleCandido said and if you want to see the two leading ones, see [Fig. 4.4c+d of my thesis](https://arxiv.org/pdf/1910.01536.pdf) of Fig. 2 of the FONLL (journal) paper.

Now, these coefficient functions are [provided in APFEL via `hqcoeff.f`](https://github.com/scarrazza/apfel/blob/bcfcd5ab49ca5d302f9bc10ce0ab70e7076664e6/src/DIS/hqcoef.f#LL144C38-L144C38), but they doesn't seemed to be used ... however, we know for sure that these terms play an important role in the FONLL prescription, because in order to work you need Eq. (71) of the FONLL paper where the "missing" terms explicitly enter. If I remember correctly, APFEL uses Eq. (103) instead of Eq. (97) where $L_q$ is "missing".

In the specific commit you mention I just disentangled the heavy and intrinsic contributions, which now we can (but before we couldn't). (I'm not even sure, that before we were doing fully consistent.) Note, that the "missing" (I think you're talking about)
https://github.com/NNPDF/yadism/blob/a182c869fa399c5d8715760fb44272d1dc150270/src/yadism/coefficient_functions/__init__.py#L155-L156
is guarded by a for loop and we're iterating on the higher quarks (this is related to #202).

PS: I don't understand when github is showing some snippets and when not ...

RoyStegeman:
I was indeed referring to the line you linked

> Now, these coefficient functions are [provided in APFEL via hqcoeff.f](https://github.com/scarrazza/apfel/blob/bcfcd5ab49ca5d302f9bc10ce0ab70e7076664e6/src/DIS/hqcoef.f#LL144C38-L144C38), but they doesn't seemed to be used ...

I see. My question was motivated by the fact that without the inclusion of those missing terms Apfel and Yadism agree to similar degree for both FFNS and FFN0, but with these terms there is significant disagreement for FFN0. Your explanation seems to make sense so if you are confident this is correct I'll just add a comment to that effect (in case people get confused while benchmarking int he future) and we accept this solution. 

If we accept this, then I think this PR is almost done? I'll check again if everything is collected as intended and do a final cleaning, but other than that I believe this is done.

felixhekhorn:
> Looks like @felixhekhorn already took care of removing the dead code (I was thinking of the old matching classes), so if everyone is happy I think the PR is done (up to maybe some minor things).

yeah, it was [confusing me](https://github.com/NNPDF/yadism/pull/195/commits/983008b7b84f92f18a2a6a3e61c54c1c9a6c6527) :upside_down_face: 

> Should it wait for [NNPDF/pineko#99](https://github.com/NNPDF/pineko/pull/99) before merging?

yes, please wait ...
- if you can take care of #203
- and before merging I'd definitely like to see a comparison Numerical vs Analytical FONLL first

AleCandido:
> > Should it wait for [NNPDF/pineko#99](https://github.com/NNPDF/pineko/pull/99) before merging?
> 
> yes, please wait ...

For me, it's not mandatory. We could make a tag on current `master` and even merge, as soon as enough reviewers approve.

> * if you can take care of #203

In a separate PR, if possible. Just to make reviews easier :)

> and before merging I'd definitely like to see a comparison Numerical vs Analytical FONLL first

We can definitely compare against `master`, but for this we either need to wait NNPDF/pineko#99, or make a one-time hack for an effective multi-flavored PDF set (`partons` on its way, but definitely not close to completion).

AleCandido:
> This I'm not sure I follow. Wouldn't the way to check simply be producing fktables for the DIS datasets using numerical fonll and seeing if the predictions agree with theory 200 or 400? What do you mean by "multi-flavor" evolution/PDF set?

*Analytic* FONLL makes use of (e.g.) 4-flavors evolved PDFs to make a calculation effectively involving 3 and 4-flavors. This is done by the matching terms that have just been removed.

*Numerical* FONLL replaces these terms with actual evolution. But to make the calculation you need PDFs evolved both in 3 and 4-flavors, at the same scales.

Usually, we avoided these evolution problems in yadism by using pre-evolved LHAPDF sets. If possible, even by using PDF sets that are not evolving at all.
However, this is definitely not possible here, because the former implementation has some evolution built-in (the matching), that now should be replaced by explicit evolution.

AleCandido:
> I was indeed planning to do it at the level of fktables at once since I expect any error (at least ones that are harder to spot) to be on the yadism side. Alternatively of course I could use the PDFs of nnpdf40 that have been evolved with different `NfFF` and write a simple script to do fonll if that's what you have in mind?

Exactly :)

In the end, one way or the other the operation is the same (start from the same boundary condition, evolve, apply the grid), but associating the one way or the other makes the difference at iteration level. In principle, you'd just need to evolve once.

However, it's true that you're currently working on FK tables support, and it might be not that far. While the proper PDF grids support is late to come, so this could only be a hack.

RoyStegeman:
I'll just do it with yadism isolated, i.e. do the check you proposed. It should be a quick check and either we immediately find something is wrong or if not we can forget about this and move on.

RoyStegeman:
Do you know how `NNPDF30_nnlo_as_0118_nf_4` and `NNPDF30_nnlo_as_0118_nf_4` have been generated? I've been using them to test the FONLL implementation assuming they only differ in evolution, can anyone confirm?

Essentially what I'm doing is as below for numerical fonll (starting from theory 400) and then comparing it to the master branch results generated for the standard theory 400 applying the pdf with `nf=4`.

There is some agreement but in particular for low Q2 where FFN03 and FFNS3 terms don't simply cancel I find disagreement up to 30% for the observable and kinematics I've checked. 

```
theorycards = [copy.deepcopy(theorycard) for _ in range(3)]
theorycards[0]["FNS"] = "FFNS"
theorycards[0]["NfFF"] = 3
theorycards[1]["FNS"] = "FFN0"
theorycards[1]["NfFF"] = 3
theorycards[2]["FNS"] = "FFNS"
theorycards[2]["NfFF"] = 4

pdfnames = [pdf_nf3, pdf_nf3, pdf_nf4]

values = []
for pdfname, tc in zip(pdfnames, theorycards):
    pdf = lhapdf.mkPDF(pdfname)
    out = yadism.run_yadism(tc, observablecard)
    values.append(out.apply_pdf(pdf))
```

giacomomagni (commented):
I left some minor comments.
It's possible to clean up the benchmark  runners (´sandbox, apfle_´ ...) a bit further but this it's fine.

src/yadism/coefficient_functions/asy/f3_nc.py
```diff
@@ -0,0 +1,34 @@
+from ..intrinsic import f3_nc as intrinsic
+from ..light import f3_nc as light
+from ..partonic_channel import EmptyPartonicChannel
+from . import partonic_channel as pc
+
+
+# TODO extract coefficient function from Buza/LeProHQ: see Eq. B.4 of Nucl. Phys. B485-420
+# and remember that in the non-singlet case d_{2xg1} = d_{xF3}
```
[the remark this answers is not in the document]
giacomomagni: > To remember: since now we do have `g1_asy` implemented we can be able to address this @felixhekhorn

src/yadism/input/domains.yaml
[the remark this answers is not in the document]
RoyStegeman: Is there a reason it isn't boolean already? Looks like it's always used that way even before.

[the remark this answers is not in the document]
AleCandido: It was like that in APFEL, (despite FORTRAN 77 having booleans, called [`LOGICAL`](https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vna9/index.html)s).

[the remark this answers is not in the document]
AleCandido: However, I will drop this file asap. You could even detach the inspector immediately, and stop maintaining it.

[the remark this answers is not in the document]
RoyStegeman: Thanks, assumed it was like that. I'll add it to #198 instead of this PR.

src/yadism/coefficient_functions/asy/f3_nc.py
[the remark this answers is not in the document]
felixhekhorn: yes! if you have time, please open a PR

src/yadism/coefficient_functions/__init__.py
```diff
@@ -69,7 +69,7 @@ def light_component(self):
         comp.extend(light.kernels.generate(self.esf, nf))
         # instead missing encodes mass effects and so we need to fork:
         for ihq in range(nf + 1, 7):
-            if masses[ihq]:
+            if not masses[ihq]:
```
AleCandido, line 72: This I believe to be plainly wrong :)

However, since we are computing single observables, and they have a single mass (we always said we are not taking into account 2-masses effects, i.e. two different massive particles in the same diagram), we should stop iterating `ihq`. We can just specify the `ihq` of the massive particle, and the numerical value of its mass, as inputs in the runcard.

Before, the motivation for iterating was exactly the FONLL implementation. But now that FONLL is not any longer in `yadism`, we should stop doing it.

RoyStegeman: Yes you're right, this changes the FFNS result which should not be changed.

For some reason the `heavy.kernels.generate_missing` and `asy.kernels.generate_missing_asy` don't agree in the large-Q limit so I'm still trying to understand why that is.

AleCandido: > For some reason the `heavy.kernels.generate_missing` and `asy.kernels.generate_missing_asy` don't agree in the large-Q limit so I'm still trying to understand why that is.

I guessed it was for this reason, as the previous commits, but it was worth to point out that this couldn't work, even to avoid you wasting too much time on this option.

RoyStegeman:
@felixhekhorn since you're the expert, do you have any thoughts on how the `missing_asy` function could be bugged? Otherwise there are no clear unexpected behaviours left in this PR that I am aware of.

felixhekhorn:
> @felixhekhorn since you're the expert, do you have any thoughts on how the `missing_asy` function could be bugged? Otherwise there are no clear unexpected behaviours left in this PR that I am aware of.

Mmm ... something *is* working https://github.com/NNPDF/yadism/pull/195/commits/40ad452f9fe8d50c5ae984cea1894e52d36aa437 and that something is $F_L$ ... for $F_2$ it is not ... I will need to dig a bit

AleCandido:
> > I would have said that the comparison should be done always at observable level...
> 
> I was, I just assumed that FFN03 and FFNS3 should cancel in the large-Q limit which @felixhekhorn now suggests may be an incorrect assumption for the missing terms. To test if the current implementation is correct even if FFNS3 and FFN03 do not cancel I'll first have to produce nf=3 and nf=4 PDFs that have been evolved consistently for both PDF and alphas.

I believe @felixhekhorn was arguing that *missing* will not cancel on its own, between FFNS3 and FFN03, but I'm pretty sure it should cancel at the observable level.

If it doesn't, then the whole FONLL has an issue (since in the large $Q^2$ limit should coincide with ZM4).

felixhekhorn:
with https://github.com/NNPDF/yadism/pull/195/commits/c6efb2bbee8c23e4eb2c3dfe1d7160ee1c54a2e9 the missing asy tests are running and the conclusion is: we were hunting shadows, aka float numerics.
- my coefficient functions are good as is and >5 years ago I already checked them analytically against Buza 
grafik
grafik
and that works out (I just forgot)
- we were not passing before, because this hack https://github.com/NNPDF/yadism/blob/c6efb2bbee8c23e4eb2c3dfe1d7160ee1c54a2e9/src/yadism/coefficient_functions/heavy/f2_nc.py#L137 was too restrictive - and we're still only just passing in the small x region (instead for the rest we match better)
- and concerning the virtual diagram we thought was missing in yadism: no, it is here: https://github.com/NNPDF/yadism/blob/c6efb2bbee8c23e4eb2c3dfe1d7160ee1c54a2e9/src/yadism/coefficient_functions/heavy/f2_nc.py#L146 (virtual ~ integral ~ quad) (so, not in my thesis, but still in yadism)
- the hack is neeeded, because of float numerics breaking down: the $d_q$ are implemented analytically in LeProHQ (see section D.6 of my thesis) - however, since they contain some nasty logs and dilogs float breaks down and you get NaNs ... so the "correct" fix would be to improve LeProHQ and to use different analytical expression in different region of kinematics

all in all missing coeff fnc should be fine, further problems still might be related to the kernels, which by the way also got adjusted here https://github.com/NNPDF/yadism/pull/195/commits/f1f82a6dedf2c92b00740dff26bf15f3f12657ca

src/yadism/coefficient_functions/asy/f3_nc.py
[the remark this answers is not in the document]
felixhekhorn: https://github.com/NNPDF/yadism/pull/209

src/yadism/runner.py
```diff
@@ -146,7 +146,7 @@ def __init__(self, theory: dict, observables: dict):
             GF=theory["GF"],
             M2W=theory["MW"] ** 2,
             M2target=theory["MP"] ** 2,
-            fonll_parts=theory["fonll-parts"],
+            fonll_parts=theory["FONLLParts"],
```
andreab1997, line 149: I believe you should call this `fonllparts` as in `coefficient_functions/__init__.py`

RoyStegeman:
I was thinking about the definition of the FFNS scheme again, and wondering whether we really even need the distinction between FONLL-FFNS and "regular" FFNS. Since FFNS already existed in yadism I did not want to change it in this PR, but depending on the use case (which I'm not even sure there is within NNPDF at least) it may be fine to define FFNS as just having a single massive quark. I.e. I am proposing to replace FFNS with what is now called FONLL-FFNS so there are not two definitions floating around.

Perhaps FFNS is already something well-known and well-defined in literature (e.g. due to apfel) and people expect it to include all massive quarks in which case we should probably keep it as it is now, but I was just wondering what you think @felixhekhorn @AleCandido.

AleCandido:
So: I was initially puzzled exactly because there is hardly anything else than FFNS at all.

Calculations are available for a given number of light flavors and zero or one massive quarks, and the asymptotic limit. Everything else is a combination.
That's why we (or at least I) wanted to suppress FONLL, ZM-VFNS and everything similar to that, at least at low level.

The only spurious bit is `missing`, since that one  is contribution to light, but you may have multiple heavy quarks contributions to missing (e.g. charm and bottom) even excluding simultaneous heavy quarks effects.

All in all, I believe we should just expose the low-level features:
- number of light flavors
- which one heavy quark (if any)
- in case of heavy, whether finite or asymptotic
- which heaviness, splitting light from missing (i.e. light, missing, charm, bottom, top), I would even avoid total at this level

Then, we could have higher level wrappers for the combinations, that at this I would at the level of grids.
Notice that this is effectively zero cost: different contributions are separately integrated by yadism anyhow. The extra cost is only dumping multiple grids and merging, that is usually negligible wrt integration time.

In this way we would have a better chance to avoid doubling, because the places where the combinations happen will be more clearly decoupled (with a different level of technicality, while now everything is more or less intertwined).

RoyStegeman:
@felixhekhorn could you check if you're happy with the last commit? If so I'll also have to do something similar for the other structure functions.

src/yadism/coefficient_functions/partonic_channel.py
```diff
@@ -150,6 +150,27 @@ def decorator(self, f):
         """
         return f
 
+    def is_below_threshold(self, z):
```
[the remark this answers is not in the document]
felixhekhorn: I know, but you lifted from `NeutralCurrentBase` (which provided the needed scope) to global ... for CC the threshold is implicit again (as for IC) provided by the convolution point (I think): https://github.com/NNPDF/yadism/blob/337758cee574d630a5d7d11b4bec26f61b64ebab/src/yadism/coefficient_functions/heavy/partonic_channel.py#L100

src/yadism/coefficient_functions/asy/partonic_channel.py
```diff
@@ -155,3 +155,24 @@ def decorator(self, f):
         if self.is_below_pair_threshold(self.ESF.x):
             return lambda: pc.RSL()
         return f
+
+    def is_below_pair_threshold(self, z):
```
felixhekhorn, line 159: since they are scoped back under NC by the parent class name you could (or not) remove the `_pair` again :upside_down_face:

RoyStegeman: Yes I could, but I think this name is also fine

AleCandido (commented):
It seems like I forgot a pending comment, but I don't remember which one...

RoyStegeman:
@AleCandido @felixhekhorn @giacomomagni This is ready for review (hope I didn't forget anything), so if any of you has some time it would be great if you could have a look.

src/yadism/coefficient_functions/asy/kernels.py
```diff
@@ -0,0 +1,191 @@
+"""Collections implementing the asymptotic components for FONLL prescription.
+
+This is strictly following the original reference :cite:`forte-fonll`, and
```
giacomomagni, line 3: Here we need to fix some docstrings.
At least the more confusing ie the one citing the original fonll paper.

src/yadism/coefficient_functions/__init__.py
```diff
@@ -27,14 +26,13 @@ def __repr__(self):
 
 
 class Combiner:
-    """
-    Does the matching between coefficient functions and partons with their approptiate coupling
-    strength.
+    """Does the matching between coefficient functions and partons with their approptiate coupling strength.
```
giacomomagni, line 29:
```suggestion
    """Does the matching between coefficient functions and partons with their appropriate coupling strength.
```

giacomomagni (commented):
* Hopefully after the benchmarks the implementation is correct. 
* There are some docstrings which can be update
* I think this comment is still valid https://github.com/NNPDF/yadism/pull/195#discussion_r1233658578, ie `skip_heavylight` is not used anymore (but please double ckeck).
* Benchmark runners are quite dirty/contain old stuff, but this should be fine.

RoyStegeman:
@giacomomagni thanks for your comments, I hope I addressed all of them adequately. 

> Benchmark runners are quite dirty/contain old stuff, but this should be fine

It was quite dirty when I found it. However, since FONLL won't be benchmarked against apfel again (at least using this runner) I'd also be happy to undo the changes to the `apfel_.py` module introduced by this PR.

AleCandido:
> It was quite dirty when I found it. However, since FONLL won't be benchmarked against apfel again (at least using this runner) I'd also be happy to undo the changes to the `apfel_.py` module introduced by this PR.

We could stop benchmarking FONLL, since it requires quite different operations. But it might be worth to benchmark instead `FFNO`, that was in APFEL, but not in yadism before this PR.

src/yadism/coefficient_functions/asy/kernels.py
```diff
@@ -0,0 +1,186 @@
...
+        km = kernels.Kernel(
+            weights["ns"], asy_cfs.__getattribute__(name)(esf, nf, m2hq=m2hq)
+        )
+        asys.append(km)
+    return asys
+
+
+def generate_heavy_asy(esf, nf, pto_evol, ihq):
+    """
+    |ref| implements :eqref:`91`, :cite:`forte-fonll`.
```
giacomomagni, line 67:
```suggestion
     Collect the high-virtuality limit of heavy.
```

src/yadism/coefficient_functions/asy/kernels.py
```diff
@@ -0,0 +1,186 @@
...
+                                asy_weights[f"{c}{av}"],
+                                asy_cfs.__getattribute__(name)(esf, nf, m2hq=m2hq),
+                            )
+                        )
+    return asys
+
+
+def generate_intrinsic_asy(esf, nf, pto_evol, ihq):
+    """
+    |ref| implements :eqref:`10` of :cite:`nnpdf-intrinsic`.
```
giacomomagni, line 124:
```suggestion
     Collect the high-virtuality limit of intrinsic, see also  :eqref:`10` of :cite:`nnpdf-intrinsic`.
```

tests/yadbox/test_export.py
```diff
@@ -17,6 +17,8 @@ def test_pineappl_sf(tmp_path: pathlib.Path):
         "F2_light": [{"x": 0.1, "Q2": 10.0}],
         "F2_total": [{"x": 0.2, "Q2": 20.0}],
     }
+    # TODO: move this to banana
```
felixhekhorn, line 20: let's remember that TODO

RoyStegeman: Thanks for pointing it out! This is no longer needed: https://github.com/NNPDF/yadism/blob/bb150a1cf509711ddc209bc579cbca0cda65a08a/src/yadism/input/compatibility.py#L146

andreab1997:
I realized something (tell me if you agree): when we ask yadism to do renormalization scale variations to the 7 subgrids of nFONLL for FONLL-B, it is using a different `PTO` for each of them (either 1 or 2). This means that, for some of them, we are doing actually NNLO ren sv instead of NLO (as we used to do in aFONLL and as we should do I believe). @RoyStegeman Can you confirm this? 

CC: @felixhekhorn, @giacomomagni @AleCandido

src/yadism/esf/esf.py
```diff
@@ -116,11 +116,12 @@ def compute_local(self):
         if self._computed:
             return
         cfc = cf.Combiner(self)
-        full_orders = [(o, 0, 0, 0) for o in self.orders]
         # prepare scale variations
         sv_manager = self.info.configs.managers["sv_manager"]
         if sv_manager is not None:
             full_orders = sv.build_orders(self.info.configs.theory["pto"])
```
felixhekhorn, line 122: as @andreab1997 pointed out: this should be `"pto_evol"`. Since FONLL-B is a NLO scheme we also need to have NLO scale variations. Actually, this means that also the `massive` grids need to have that setting!

andreab1997:
> Thanks for pointing this out - I did not think about SV, but indeed if SV is NLO for all contributions (i.e. similar to FONLL-A) instead of the same order as FONLL-B for the central scale, then for SV the runcards will need to be updated to have PTO=1. If it were only FFN0 it would indeed have been possible to replace `pto` with `pto_evol`, but for massive this won't work.
> 
> So I suppose this should be fixed in pineko?

I believe at this point we should just have a `PTO` (we can call it as you like) that we use for `EKO` and for the renormalization scale variations in `yadism` and that would be always 1 for FONLL-A and B and always 2 for FONLL-C and D. Then we need another PTO (that can be PTODIS maybe) that instead is used only by `yadism` and that is 2 always for FONLL-C but it is the usual 2,2,1,2,2,1,2. Then we would be only left with what `PTODIS` is used at the moment: as far as I understood it is used for the FFNS0. Could this be lifted to yadism?

RoyStegeman:
Your proposal is exactly what yadism already does for the central scale, though I'm not exactly sure what you mean here: 
> Then we would be only left with what PTODIS is used at the moment: as far as I understood it is used for the FFNS0. Could this instead be lifted to yadism?

I'd have to look at how SV is implemented to see if it requires any changes, but from what I understood SV at NLO corresponds to both PTO=1 and PTODIS=1?

andreab1997:
> Your proposal is exactly what yadism already does for the central scale, though I'm not exactly sure what you mean here:
> 
> > Then we would be only left with what PTODIS is used at the moment: as far as I understood it is used for the FFNS0. Could this instead be lifted to yadism?
> 
> I'd have to look at how SV is implemented to see if it requires any changes, but from what I understood SV at NLO corresponds to both PTO=1 and PTODIS=1?

As far as I understood PTODIS is used for the `asy` parts in the case of FONLL-B. For SV it would be enough to ask for the sv kernel related to PTO=1 for each of the 7 theories, which in my mind could have been `PTO` since it has to be 1 already for the eko computation. So I would say SV should not use PTODIS at all. What I was saying is that if `PTO` we put equal to 1 always (for the reasons I said), then in yadism you need a different parameter that should be 2,2,1,2,2,1,2, according to the subtheory (which at the moment is instead PTO). Right?
As far as I understood then `PTODIS` is on top of all this discussion because being related to the `asy` components is yet another parameter which we may need to keep or to lift in yadism (if possible). But maybe I misunderstood the role of `PTODIS`

felixhekhorn:
Just to be clear (and as @RoyStegeman already pointed out, it can be confusing): there are only two (2) couplings powers in the game - and no more. Unfortunately there names are ambiguously:
1. the (old) "PTO" in the theory card: since this is inherited from a NNPDF theory this refers to the PTO of evolution and in yadism this corresponds thus to `pto_evol`
2. the (new) "PTODIS" in the theory card: this we invented ourselves to exactly accommodate FONLL-B and it refers to `pto` inside yadism

the fallback for 2 is 1

- 2 should be used to compute coefficient functions
- 1 should be used to compute anything related to resummation, i.e. the logs in asy and $a_s$

AleCandido:
As @felixhekhorn said, `PTODIS` has been introduced to support FONLL-B (and all the skewed FONLL schemes), the whole of it, not just `asy` (indeed the fixed order in the asymptotic has to be the exact same as in the massive part, you can play just with the logs).

Concerning SV, I'm not sure I understand the point: they should be computed with the exact same settings of the central one, just varying the scale.
If central uses a non-trivial `PTODIS`, then SV have to do it as well.

felixhekhorn:
> Concerning SV, I'm not sure I understand the point: they should be computed with the exact same settings of the central one, just varying the scale.
> If central uses a non-trivial `PTODIS`, then SV have to do it as well.

The question is to which order are you generating the logs? (inside the coeff fnc <=> inside the as expansion)

src/yadism/coefficient_functions/heavy/f2_nc.py
```diff
@@ -127,26 +125,31 @@ def cq(z, _args):
...
...08037,-0.418858709662556,-0.506260665515978,-0.610996028067369,-0.736251187872195,-0.885730857584996,-1.06372519245996,-1.27518151164027,-1.52578000909733,-1.82201262156352,-2.17126399703945,-2.58189329443875,-3.06331535728638,-3.62607965724226,-4.28194531083085,-5.04395044694658,-5.92647425143555,-6.94529014126800,-8.11760872183633,-9.46210944844544,-10.9989601749240,-12.7498244246557,-14.7378558506660,-16.9876807198927,-19.5253688839296,-22.3783939650644,-25.5755843093813,-29.1470657060820,-33.1241861478952,-37.5395053973696,-42.4266052184745,-47.8201213196923,-53.7556644942167,-60.2696464182845,-67.3993063275125,-75.1826378056694,-83.6582241870451,-92.8652810526019,-102.843541628920,-113.633210752990,-125.274912395365,-137.809641501730,-151.278720093253,-165.723757499034,-181.186614539308,-197.709371436022,-215.334299216791,-234.103834361192,-254.060556378063,-275.247168049343,-297.706478254068,-321.481386973778,-346.614871863184,-373.149977148964,-401.129803760321,-430.597497710567,-461.596255112282,-494.169300847881,-528.359882503872,-564.211281969937,-601.766798840314,-641.069750271690,-682.163468181970,-725.091296878593,-769.896591054024,-816.622714094366,-865.313036654766])
```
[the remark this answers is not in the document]
felixhekhorn: done in [560dea0](https://github.com/NNPDF/yadism/pull/195/commits/560dea040d59a97521234a4d88980c1636ee2e56) - together with an improved interpolation and cut-off limit; @RoyStegeman or @niclaurenti could one of you please check this is working?

felixhekhorn (approved):
I didn't check everything super carefully, but all in all it looks good (and I trust in the benchmarks ;-) )

niclaurenti:
Accidentally pushed here...sorry

PR #207, opened 2023-06-21
Fix PyPi deployment

Radonirinaunimi:
The following fixes the error causing the [deployment](https://github.com/NNPDF/yadism/actions/runs/5332480725/jobs/9661710446#step:11:34) to fail. I am not 100% certain that `-E mark` is also needed in which case we could also remove it.


AleCandido:
> The following fixes the error causing the [deployment](https://github.com/NNPDF/yadism/actions/runs/5332480725/jobs/9661710446#step:11:34) to fail. I am not 100% certain that `-E mark` is also needed in which case we could also remove it.

Unit tests span also `yadmark` and `yadbox`, that's why you need PineAPPL in the first place.
So, yes, but I'll immediately convert the `pineappl` extra into `box` :)

AleCandido:
Like this it should work, and be closer to EKO :)

AleCandido (approved):

Radonirinaunimi:
Perfect! Thanks @AleCandido, I will regenerate the `poetry.lock` and we should be good to go.

AleCandido:
I can not regenerate immediately the lock file, because yadism doesn't support py3.11, so Poetry is complaining. I'm doing in a container, just wait a second...

felixhekhorn (approved):

AleCandido:
> it should - I added the support just a few days ago: [a2f81d3](https://github.com/NNPDF/yadism/commit/a2f81d3fad74aec134ae13084637eda3ec2418c9) (also see the CI)

Ok, then you made a tiny mistake:
https://github.com/NNPDF/yadism/blob/1fb178e495be30c6f31976b499393938871e8943/pyproject.toml#L39
should be:
```toml
 python = ">=3.8,<3.12"
```
(indeed before it was `<3.11`).

The problem is that py3.11.2 is not `<=` py3.11

AleCandido:
@Radonirinaunimi now it should work, feel free to merge whenever you wish :)

PR #208, opened 2023-06-21
Fix pydocstyle

Radonirinaunimi:
It was strongly advised before to fix the pydocstyle in the same way as what was done in https://github.com/NNPDF/eko/pull/243 and hence the reason for this PR. I volunteered to do so and will do it incrementally, but everyone is invited to contribute.

- [ ] yadism
  - [ ] coefficient_functions
  - [ ] esf
  - [ ] input
- [ ] yadbox
   - [ ] runcards 
- [ ] yadmark
  - [ ] benchmark
  - [ ] data
  - [ ] navigator


AleCandido:
Thanks @Radonirinaunimi, it's an annoying task, but useful.

However, since part of the task really consists in reviewing the docstrings, I would suggest you not to target (necessarily) completion. Even reducing the amount of errors might be enough, and a significant improvement.

Radonirinaunimi:
> Thanks @Radonirinaunimi, it's an annoying task, but useful.
> 
> However, since part of the task really consists in reviewing the docstrings, I would suggest you not to target (necessarily) completion. Even reducing the amount of errors might be enough, and a significant improvement.

Yes, indeed! That'd be exactly the plan.

felixhekhorn:
@Radonirinaunimi do you want to merge what is already here? just to not loose it? and convert the list into an issue?

Radonirinaunimi:
> @Radonirinaunimi do you want to merge what is already here? just to not loose it? and convert the list into an issue?

My personal goal is to finish this before the Yadism paper appear. If that is not achieved, I agree that we can merge already what is here. But I haven't said my last words yet 😄

felixhekhorn:
better push for the paper itself :innocent: or if you want to code, go for benchmarks

Radonirinaunimi:
> better push for the paper itself 😇 or if you want to code, go for benchmarks

Absolutely! But this, if it will happen, will be for the outside hours.

PR #215, opened 2023-08-01
Approximate massive N3LO for F2 and FL

giacomomagni:
This PR can introduce the approximate massive N3LO contribution for `F2,FL` from @niclaurenti thesis and the relative code:  https://github.com/niclaurenti/adani

For the time being: 

* Asymptotic coefficient functions can be computed directly using `adani`. 
* Massive coefficient functions have to be interpolated from a precomputed grid, since their computation is too slow.
* A new theory key is needed to select the possible coefficient function variation


giacomomagni:
@niclaurenti to implement the asymptotic limit needed in FONLL, 
following the yadism strategy we need to provide the high-scale (`high-scale = Q^2 >> m^2` in `adani` convention) coefficients split by logs $L=log(Q^2/m^2)$ with $L_\mu = L_Q$.
This requires some further manipulation of the expression `DX_yn_highscale` given in `adani` right?  
Any suggestion how to proceed ?

niclaurenti:
> * Massive coefficient functions have to be interpolated from a precomputed grid, since their computation is too slow.

Hi @giacomomagni , for the approximation are you using the argument `method_flag=1`? (if not provided it will use 0).
Since `method_flag=1` uses monte carlo integration and it is much faster than `method_flag=0`.

niclaurenti:
> @niclaurenti to implement the asymptotic limit needed in FONLL,
> following the yadism strategy we need to provide the high-scale (`high-scale = Q^2 >> m^2` in `adani` convention) coefficients split by logs  with .
> This requires some further manipulation of the expression `DX_yn_highscale` given in `adani` right?
> Any suggestion how to proceed ?

Inside `adani` it is not possible to get only the contribution proportional to $\log Q^2/m^2$.
The only thing that I can think of is to reimplement the expressions splitting the terms. It will not be difficult but only very long (given that the expressions are huge)

niclaurenti:
> @niclaurenti to implement the asymptotic limit needed in FONLL,
> following the yadism strategy we need to provide the high-scale (`high-scale = Q^2 >> m^2` in `adani` convention) coefficients split by logs  with .
> This requires some further manipulation of the expression `DX_yn_highscale` given in `adani` right?
> Any suggestion how to proceed ?

Inside `adani` it is not possible to get only the contribution proportional to $\log Q^2/m^2$.
The only thing that I can think of is to reimplement the expressions splitting the terms. It will not be difficult but only very long (given that the expressions are huge)

giacomomagni:
Hi @niclaurenti, yes also using `method_flag=1` was quite slow, but I hope we can do a good job also using the grids.

Regarding the `high-scale` limit, that is mandatory to implement FONLL, so it has to be separated...

AleCandido:
It would be helpful to know what is $\mu$.

Concerning point 2.: we're (or I'm) planning to move yadism coefficient functions to a compiled extension (sooner or later). Making the effort of coding them with Numba, to possibly undo it seems not worth it.
On the other hand, yadism has no compiled extension, for the time being. I would keep them in `adani` for a while, or in case move them to `LeProHQ` (to have a single other provider, but it should not change a lot), and migrate whenever we'll start implementing #219 (if ever).

niclaurenti:
> It would be helpful to know what is μ.

It's the factorization scale: those functions have a log dependence on both $Q$ and $\mu$ (see for example arXiv:1403.6356v1 from eq. 597)

> 
> Concerning point 2.: we're (or I'm) planning to move yadism coefficient functions to a compiled extension (sooner or later). Making the effort of coding them with Numba, to possibly undo it seems not worth it. On the other hand, yadism has no compiled extension, for the time being. I would keep them in `adani` for a while, or in case move them to `LeProHQ` (to have a single other provider, but it should not change a lot), and migrate whenever we'll start implementing #219 (if ever).

Ok perfect. I will try to do it as soon as possible

giacomomagni:
> It's the factorization scale: those functions have a log dependence on both Q and μ (see for example arXiv:1403.6356v1 from eq. 597)
> 

I'd agree to set $\mu_f = Q$ and then split massive logs. 

Regarding point 2) let's keep them in `adani` is possible also because here we might not have all the polylogarithms needed..

AleCandido:
The renormalization/factorization scale contributions are managed separately, you never need to take care of them in the coefficient functions (for yadism, in adani to whatever you wish).
Sooner or later they might also fully migrate to Pineko, cf. #220

giacomomagni:
This should be now ready for review.

felixhekhorn (approved):

PR #222, opened 2023-08-21
Fix polarised TMC

Radonirinaunimi:
The following fixes bugs present in the computation of polarised TMC. The bugs originate from the fact that our definition of `g1` (as called by the line below):

https://github.com/NNPDF/yadism/blob/9e8d44568ae766e241da8ba17aba3b4809b97134/src/yadism/esf/tmc.py#L551

is actually $2xg_1$ and therefore the integrands were wrong. A Mathematica notebook is added that contains the correct integrals and computes their approximations.


giacomomagni (approved):

src/yadism/esf/tmc.py
```diff
@@ -557,48 +535,73 @@ class ESFTMC_g1(EvaluatedStructureFunctionTMC):
...
+        return factors * g1out
 
     def _get_result_exact(self):
-        # collect g1 results
+        # Collect g1 result.
         g1out = self.sf.get_esf(self.sf.obs_name, self._shifted_kinematics).get_result()
+
         # Call to the raw integrals
-        k1out = self._k1()
+        k1out = self._h2()
```
felixhekhorn, line 576: do you want to add an explicit redirect? else the names are a bit confusing ...

felixhekhorn (approved):
I didn't check the math explicitly, but I can easily imagine having a normalization issue around

src/yadism/esf/tmc.py
Radonirinaunimi: I can indeed make a copy of that function (although this was what I tried to avoid) to make it more explicit.

As for the normalization, as long as everything is consistent, in that we account for a factor of $1/2$ from the integrand, $1/(2 \xi)$ everywhere $g_1$ appear, and undo by $2 \xi$ in overall, then everything is correct. This indeed I cross-checked a couple of times, and I think is also verified by the comparison plots in the Appendix of the draft.

felixhekhorn: a simple redirect should be sufficient `_h2 = _k1` no?

felixhekhorn: (adding updated docstrings is a good idea - however, just to say, also that simple line `_k1 = _h2` would have worked :upside_down_face: )

PR #226, opened 2023-09-12
Fix numerical fonll

giacomomagni:
Here we try to debug nFONLL deeply such that it will reproduce exactly aFONLL.

- [x] Remove the kinematic threshold from the asymptotic coeff functions.
- [x] Check numerical accuracy of the single cancellations
- [x] Check IC contributions

### DIFF
--- a/src/yadism/coefficient_functions/asy/f2_nc.py
+++ b/src/yadism/coefficient_functions/asy/f2_nc.py
@@ -7,8 +7,8 @@ from . import raw_nc
 class AsyLLGluon(pc.NeutralCurrentBaseAsy):
     def NLO(self):
         def cg_LL_NLO(z, args):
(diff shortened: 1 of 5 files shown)


giacomomagni:
this should not be needed, because the asymptotic depends only $\ln(m^2/Q^2)$.

https://github.com/NNPDF/yadism/blob/65ab2ed2292c770baa3bbe778987dfcf6e06b670/src/yadism/coefficient_functions/asy/partonic_channel.py#L16C9-L16C13

giacomomagni:
What remain to be done is fixing the tests. 
Do you know why it's not passing anymore @felixhekhorn ?

@RoyStegeman @andreab1997

RoyStegeman:
Thanks, I'll recalculate the theory and produce a fit.

PR #233, opened 2023-10-16
Compute F1

toonhasenack:
We added a case in the yadism/src/esf/exs.py file where one would like to calculate F1. Dependencies of this F1 have also been added.

### DIFF
--- a/src/yadism/esf/exs.py
+++ b/src/yadism/esf/exs.py
@@ -66,6 +66,8 @@ def xs_coeffs(kind, y, x=None, Q2=None, params=None):
         ym = 0
         yp = 1.0
         yL = y**2 / (2 * (y**2 / 2 + (1 - y) - (mn * x * y) ** 2 / Q2))
+    elif kind == "F1":
+        return np.array([1 / (2 * x), -1 / (2 * x), 0])
     elif kind == "XSFPFCC":
         INV_GEV_TO_PB = GEV_CM2_CONV / 100.0  # pb
         norm = (INV_GEV_TO_PB * params["GF"] ** 2) / (2.0 * np.pi)
--- a/src/yadism/observable_name.py
+++ b/src/yadism/observable_name.py
@@ -9,6 +9,7 @@ xs = [
     "XSNUTEVCC",
     "XSNUTEVNU",
     "FW",
+    "F1",
     "XSFPFCC",
 ]
 kinds = sfs + xs + [fake_kind]
--- a/tests/yadism/esf/test_exs.py
+++ b/tests/yadism/esf/test_exs.py
@@ -50,6 +50,17 @@ def test_xs_coeffs():
     assert fw[0] == 1.0
     assert fw[2] == 0.0
 
+    f1 = xs_coeffs(
+        "F1",
+        x=0.5,
+        Q2=1.0,
+        y=0.0,
+        params=dict(projectilePID=1, M2target=0.0, M2W=1.0),
+    )
+    assert f1[0] == 1.0
+    assert f1[1] == -1.0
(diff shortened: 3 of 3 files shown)


Radonirinaunimi:
[621d32f](https://github.com/NNPDF/yadism/pull/233/commits/621d32fee95fb68628a27f2d67e24303d0083602) should not be part of this PR. Could you please remove it?

felixhekhorn (approved):

Radonirinaunimi (approved):

felixhekhorn:
@Radonirinaunimi actually, is this consistent? When you request the observable `g1` from yadism you get $2xg_1$ as inherited from LeProHQ, right? - see also https://github.com/NNPDF/yadism/pull/222 ... shouldn't we do here the same? i.e. you request `F1` but you get $2xF_1$? it is a bit confusing, I know, but if we do the same thing in polarized and unpolarized we release some of the tension, because all that matters is $g_1/F_1 = (2xg_1)/(2xF_1)$

giacomomagni:
> it is a bit confusing, I know, but if we do the same thing in polarized and unpolarized we release some of the tension, because all that matters is g1/F1=(2xg1)/(2xF1)

Good point, maybe the best would be the other way round ie remove the `2x` pre-factor also from `g1`, no ?

felixhekhorn:
> Good point, maybe the best would be the other way round ie remove the `2x` pre-factor also from `g1`, no ?

this is of course an equally good option - the reason why I chose $2xg_1$ in my thesis is that with this normalization $F_2$ and $2xg_1$ have the analogous parton model expression, see e.g. PDG
grafik
(same holds for $xF_3$ ... )

Radonirinaunimi:
You are absolutely right @felixhekhorn! The initial idea was to account for the extra $2x$ in the $g_i$ later and keep using $F_1$. In hindsight, since the $F_1$ is mainly relevant in the $g_1/F_1$ context, we should provide here $2xF_1$ instead.
```python
    elif kind == "2xF1":
        return np.array([1, -1, 0])
```

Radonirinaunimi:
I think this would be the easiest option (taking also into account your physics' argument), so if you agree @felixhekhorn, @giacomomagni we can correct this PR according to that.

giacomomagni:
> ```python
>     elif kind == "2xF1":
>         return np.array([1, -1, 0])
> ```

But let's keep calling it F1, otherwise it's even more inconsistent with the rest (right now both `g1` and `F3` are multiplied by x). 
We should edit the documentation and spell this out properly there.

toonhasenack:
Okay, if everyone agrees I will generate a new pull request

felixhekhorn:
> But let's keep calling it F1, otherwise it's even more inconsistent with the rest (right now both `g1` and `F3` are multiplied by x). We should edit the documentation and spell this out properly there.

I perfectly agree this should be proper documented

just to be explicit: what I believe are equal player are $F_2, F_L, xF_3, 2xF_1$ for unpolarized structure functions and $2xg_1, g_4, g_L, 2xg_5$ for the polarized case (see PDG above)

Radonirinaunimi:
> > But let's keep calling it F1, otherwise it's even more inconsistent with the rest (right now both `g1` and `F3` are multiplied by x). We should edit the documentation and spell this out properly there.
> 
> I perfectly agree this should be proper documented
> 
> just to be explicit: what I believe are equal player are F2,FL,xF3,2xF1 for unpolarized structure functions and 2xg1,g4,gL,2xg5 for the polarized case (see PDG above)

Perfect! 

> Okay, if everyone agrees I will generate a new pull request

So, @toonhasenack you could now proceed this way.

PR #239, opened 2023-11-05
Nix

AleCandido:
The usual Nix PR I'm adding to all repos.

A more detailed description can be found in https://github.com/NNPDF/pineappl/pull/245


pyproject.toml
```diff
@@ -51,15 +51,8 @@ banana-hep = { version = "^0.6.6", optional = true }
 sqlalchemy = { version = "^1.4.21", optional = true }
 
 [tool.poetry.group.dev.dependencies]
-poethepoet = "^0.10.0"
```
felixhekhorn, line 54: why? is it included somewhere else? or you want to stress "Install the CLI globally from PyPI" as said [on the poe docs](https://poethepoet.natn.io/index.html#quick-start)

felixhekhorn (approved):
I don't know if the nix stuff is working, but yadism seems fine

Radonirinaunimi:
Awesome 🚀! I will test this and report.

pyproject.toml
AleCandido: The idea is that you'd rather want to stay outside the environment, and run the commands within the environment. At least you'd try to ensure that:
- commands are run with all and only the dependencies locked
- you can freely install your own tools and everything you need (for development and not) with `pip`, without modifying the environment, and move it away from the locked state

So, `poe test`, e.g., is convenient from the outside, because the corresponding command will be run inside the environment, even you are currently outside of it.

PR #240, opened 2023-11-05
Drop inspector and related features

AleCandido:
The only module preserved in the `input` subpackage is the `compatibility` one (since I believe is heavily used).

Originally, I wanted to introduce in this PR even the new runcard format, as I did in EKO (and rename the `input` subpackage to `io`, to mirror the same structure in both packages).
However, dropping the inspector has almost no consequence, in practice, since it is barely used (and almost not even tested, since some of them were skipped).
But adding the new runcards could be a breaking change, and we'll definitely need some more time for them, with potentially limited benefits in the short term.

I might author a draft PR about the new runcards in the next days, but for sure I'd like to keep it separate from this one (that is ready to merge, once reviewed).


AleCandido:
Despite I declared this to be ready to merge, remember that is pointing to #239. Thus, we should wait for that to be merged first (or rebase on `master` before merging).

The two of them should be pretty independent (the only file touched by both is `.gitignore`, and I'm about to move that commit on the other side).

felixhekhorn (approved):

felixhekhorn:
Can we merge this? either with #239 or without ... cc @Radonirinaunimi

AleCandido:
This is separate from #239, exactly because in this way you could rebase and merge on its own.

However, #239 is not harmful (worst case it's junk), so you could even merge both without major drawbacks.

PR #242, opened 2023-11-13
Bump adani version to 0.14

Radonirinaunimi:
Bumping `adani` version to `v0.14` fixes errors when building the wheel (on some platforms) according to PEP 517.


pyproject.toml
```diff
@@ -43,7 +43,7 @@ numba = "^0.57.0"
 eko = "^0.13.4"
 pandas = "^1.3.0"
 rich = "^12.4.4"
-adani = "^0.13"
+adani = "^0.14"
```
niclaurenti, line 46: Just to make you know: I've also pushed 0.14.1 that has also mac wheels

niclaurenti (approved):

pyproject.toml
Radonirinaunimi: Thanks! Updated mine [68eb5ee](https://github.com/NNPDF/yadism/pull/242/commits/68eb5ee318cc9fc2ff986043f7f307dac38a6551) and everything still works fine, so this is very good.

AleCandido (approved):

pyproject.toml
felixhekhorn: do you want to pin v0.14.1? i.e. putting it here?

PR #245, opened 2023-11-18
Extend benchmarks

giacomomagni:
This PR is to add FFNS to Apfel++ benchmarks.

### DIFF
--- a/src/yadmark/benchmark/external/apfelpy_utils.py
+++ b/src/yadmark/benchmark/external/apfelpy_utils.py
@@ -17,6 +17,69 @@ MAP_HEAVINESS = {
 }
 
 
+def map_apfelpy_sf(init, observables, fns):
+    """Get required structure function.
+
+    Parameters
+    ----------
+    init: ap.initializers
+        Apfel++ initializers
+    observables: dict
+        observables runcard
+    fns: str
+        flavor number scheme
+
+    Returns
+    -------
+        Apfel++ structure function initalizer
+    """
+    MAP_ZM_CC = {
+        "F2m": init.InitializeF2CCMinusObjectsZM,
+        "FLm": init.InitializeFLCCMinusObjectsZM,
+        "F3m": init.InitializeF3CCMinusObjectsZM,
+        "F2p": init.InitializeF2CCPlusObjectsZM,
+        "FLp": init.InitializeFLCCPlusObjectsZM,
+        "F3p": init.InitializeF3CCPlusObjectsZM,
+    }
+
+    MAP_ZM_NC = {
+        "F2": init.InitializeF2NCObjectsZM,
+        "FL": init.InitializeFLNCObjectsZM,
+        "F3": init.InitializeF3NCObjectsZM,
+        "g1": init.Initializeg1NCObjectsZM,
+        "gL": init.InitializegLNCObjectsZM,
+        "g4": init.Initializeg4NCObjectsZM,
+    }
+
+    MAP_FFNS_NC = {
+        "F2": init.InitializeF2NCObjectsMassive,
+        "FL": init.InitializeFLNCObjectsMassive,
+    }
+    MAP_FFNS0_NC = {
+        "F2": init.InitializeF2NCObjectsMassiveZero,
+        "FL": init.InitializeFLNCObjectsMassiveZero,
+    }
+
+    if observables["prDIS"] == "CC":
+        projectile_pids = {
+            "electron": 11,
+            "positron": -11,
+            "neutrino": 12,
+            "antineutrino": -12,
+        }
+        if projectile_pids[observables["ProjectileDIS"]] > 0:
+            apfelpy_structure_functions = MAP_ZM_CC
+    else:
+        if fns == "ZM-VFNS":
+            apfelpy_structure_functions = MAP_ZM_NC
+        elif fns == "FFNS":
+            apfelpy_structure_functions = MAP_FFNS_NC
+        elif fns == "FFNS0":
+            apfelpy_structure_functions = MAP_FFNS0_NC
+
+    return apfelpy_structure_functions
+
+
 def couplings(ap, pids, proc_type, obs_name):
     """Return the corresponding coupling given a process type and an observable.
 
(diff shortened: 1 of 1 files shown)


src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -334,8 +385,13 @@ def compute_apfelpy_data(theory, observables, pdf):
...
                 tab_sf = tabulate_nc(ap, obs_name, sfobj, NQ, QMin, QMax, thrs)
 
+            # shift convolution for massive
+            eta = 1.0
+            if fns == "FFNS":
+                m_h = masses[3] if heaviness == "charm" else masses[4]
+                eta = Q2 / (Q2 + 4 * m_h**2)
             # compute the actual result
-            result = tab_sf.EvaluatexQ(x, np.sqrt(Q2))
+            result = tab_sf.EvaluatexQ(x / eta, np.sqrt(Q2))
```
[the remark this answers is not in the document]
giacomomagni: okay I copied the syntax, from apfel++ ...

AleCandido:
About the naming issue, I remember that we were often calling the `x_eval` (current name) `chi`, after some paper.
In particular, K&S, if I remember correctly, meaning that was more intrinsic-related nomenclature. But we could apply to massive as well.

However, it's just a random remark, actually `x_eval` is pretty explicit.

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -17,6 +17,69 @@
...
+            apfelpy_structure_functions = MAP_ZM_CC
+    else:
+        if fns == "ZM-VFNS":
+            apfelpy_structure_functions = MAP_ZM_NC
+        elif fns == "FFNS":
+            apfelpy_structure_functions = MAP_FFNS_NC
+        elif fns == "FFNS0":
+            apfelpy_structure_functions = MAP_FFNS0_NC
+
+    return apfelpy_structure_functions
```
AleCandido, line 80: Small improvement to the code: early returns

```suggestion
        if projectile_pids[observables["ProjectileDIS"]] > 0:
            return MAP_ZM_CC
    else:
        if fns == "ZM-VFNS":
            return MAP_ZM_NC
        if fns == "FFNS":
            return MAP_FFNS_NC
        if fns == "FFNS0":
            return MAP_FFNS0_NC
```

AleCandido: In principle, you could define the various maps in place (they are used only once), and they even have the wrong case, since they are not constants (they depend on `init`).

The other option would be to define them as top-level functions, that take `init` as a parameter and return the mapping.

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -17,6 +17,69 @@
...
+        "FL": init.InitializeFLNCObjectsMassiveZero,
+    }
+
+    if observables["prDIS"] == "CC":
+        projectile_pids = {
+            "electron": 11,
+            "positron": -11,
+            "neutrino": 12,
+            "antineutrino": -12,
+        }
```
AleCandido, line 69: Instead, this is actually a constant (and could be defined top-level).

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -3,9 +3,9 @@
 from eko import basis_rotation as br
 
 # Q2 knots specs
-NQ = 250
+NQ = 1200
 QMin = 1
-QMax = 200
+QMax = 100
```
AleCandido, line 9: Is there a reason to fix these values here, and not accepting them as input?

(moreover, if they are constants they have the wrong case)

AleCandido (commented):
Mostly potential style improvements

src/yadmark/benchmark/external/apfelpy_utils.py
felixhekhorn: this actually reveals a hole in the definition: what happens if `projectile_pids[observables["ProjectileDIS"]] > 0` yields False?

src/yadmark/benchmark/external/apfelpy_utils.py
giacomomagni: oh this was committed by mistake, no reason to change them.

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -226,55 +287,42 @@ def compute_apfelpy_data(theory, observables, pdf):
...
     # Map Yadism observables to Apfel++ Objects
-    init = ap.initializers
-    if observables["prDIS"] == "CC":
-        projectile_pids = {
-            "electron": 11,
-            "positron": -11,
-            "neutrino": 12,
-            "antineutrino": -12,
-        }
-        if projectile_pids[observables["ProjectileDIS"]] > 0:
```
giacomomagni, line 251: @Radonirinaunimi if `projectile_pids` do we crash here right ?

src/yadmark/benchmark/external/apfelpy_utils.py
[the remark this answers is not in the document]
felixhekhorn: your fix [781ed78](https://github.com/NNPDF/yadism/pull/245/commits/781ed78e1926c9d1592aa07e3831788351d526e5) actually only renames eta :upside_down_face: but I wanted to shift the division inside the `if` (sorry github would not let me suggest the right code, because he thinks there were deleted lines - I'm wondering if I should file a bug report against github ...)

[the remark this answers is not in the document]
giacomomagni: I see sorry.

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -332,10 +383,15 @@ def compute_apfelpy_data(theory, observables, pdf):
...
                     theory["XIF"],
                 )
-                tab_sf = tabulate_nc(ap, obs_name, sfobj, NQ, QMin, QMax, thrs)
+                tab_sf = tabulate_nc(ap, obs_name, sfobj, NQ, QMIN, QMAX, thrs)
 
+            # shift convolution for massive
+            x_eval = x
+            if fns == "FFNS":
+                m_h = masses[3] if heaviness == "charm" else masses[4]
+                x_eval = x / (Q2 / (Q2 + 4 * m_h**2))
```
felixhekhorn, line 392:
```suggestion
                eta = Q2 / (Q2 + 4 * m_h**2)
                x_eval = x / eta
```

felixhekhorn (approved):

src/yadmark/benchmark/external/apfelpy_utils.py
```diff
@@ -16,6 +16,69 @@
...
+    }
+    MAP_FFNS0_NC = {
+        "F2": init.InitializeF2NCObjectsMassiveZero,
+        "FL": init.InitializeFLNCObjectsMassiveZero,
+    }
+
+    if observables["prDIS"] == "CC":
+        if PROJECTILE_PIDS[observables["ProjectileDIS"]] > 0:
+            return MAP_ZM_CC
+        return ValueError(f"{observables['ProjectileDIS']} not available in Apfel++")
```
AleCandido, line 73: It's not strictly an error, but I guess you wanted 
```suggestion
        raise ValueError(f"{observables['ProjectileDIS']} not available in Apfel++")
```

AleCandido (approved):

src/yadmark/benchmark/external/apfelpy_utils.py
Radonirinaunimi: Sorry for the late reply! It doesn't come to me right away why exactly this was just passed previously (as opposed to not forced to raise an exception), to be fully certain I will test this tomorrow and check (I will also review this PR alongside).

Radonirinaunimi: I have looked into this and now everything comes back, to be more exact in the error raised I'd print out the following message instead as the current one implies that it is not possible to get it from `APFEL++`:
```python
raise ValueError(f"Benchmarks not implemented for CC {observables['ProjectileDIS']} ")
```

Radonirinaunimi (approved):
Everything looks good to me!

PR #247, opened 2023-11-28
Fix singlet channel in heavylight

felixhekhorn:
- the only relevant change is line 330
- Python `range(1,nf+1)`  will run up to and including nf
- this make the singlet part also symmetric to here https://github.com/NNPDF/yadism/blob/66fcaf9aed079f43f6a68cb6bf536a5cea6397ad/src/yadism/coefficient_functions/light/kernels.py#L111
- this only has an impact for flavoured observables at NNLO

### DIFF
--- a/src/yadism/coefficient_functions/kernels.py
+++ b/src/yadism/coefficient_functions/kernels.py
@@ -1,3 +1,8 @@
+"""Coefficient function kernels.
+
+Kernels are a combination of the coupling structure (which parton with which weight)
+and the respective 'raw' mathematical coefficient function.
+"""
 import copy
 import importlib
 from numbers import Number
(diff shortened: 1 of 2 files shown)


src/yadism/coefficient_functions/kernels.py
```diff
@@ -323,11 +321,13 @@ def generate_single_flavor_light(esf, nf, ihq):
...
-
+    ns_partons = {}
     ns_partons[ihq] = w
     ns_partons[-ihq] = w if not is_pv else -w
-    ch_av = w / (nf) if not is_pv else 0.0
-    for pid in range(1, nf):
+    ch_av = w / nf if not is_pv else 0.0  # omitting again *2/2
+    # all quarks from 1 up to and including nf form the singlet
+    s_partons = {}
+    for pid in range(1, nf + 1):
```
felixhekhorn, line 330: This is the only relevant line

giacomomagni: Okay, just to understand the ratio behind this bug.
This line was changed for the last time 2 years ago, so we missed the plus one, 
because before it was included here:

https://github.com/NNPDF/yadism/pull/195/files#diff-07c4c61a53d4e5e440331572095ea2eb341a4bcb4a3f22863449fb6021cd624eL207

felixhekhorn: Unfortunately your link is not working - but yes, the bug was always there since the `generate_single_flavor_light` function was introduced here https://github.com/NNPDF/yadism/commit/b6c352b801cd5361db1b0f20554488b905da21f6 on Jul 12, 2021; you can see also from the removed lines in that commit that before it had the +1 - however, most likely this has no visible effect (to be proven) ... I just stumbled about this while looking for a possible bug

AleCandido: I agree it's a bug, and it could have been generated by confusion over the +/- 1 in the number of light flavors (also the reason why `ihq` is named like that, and not `nhq`).

However, that's kind of historical.
Most likely the dominant reason is that it only affects very mildly the full observable (because it's a piece of a NNLO contribution for a single light quark, only used in certain situations), and it has not been benchmarked in isolation.

giacomomagni (approved):

PR #255, opened 2024-01-18
Add some tutorials

Radonirinaunimi:
For the sake of the paper https://github.com/N3PDF/papers/pull/26, this adds a short tutorial that should only contain the essentials.

As it is right now, this is not ready yet and I will be working on the contents. But are we satisfied with the structure? If so, then I can carry on.

cc @alecandido, @felixhekhorn, @giacomomagni, @RoyStegeman: if there are other aspects to be covered, please suggest and/or feel free to edit directly.


felixhekhorn:
Should we weave (copy) the Como tutorial in here?

Radonirinaunimi:
> Should we weave (copy) the Como tutorial in here?

That might be a good idea. Maybe not the whole tutorial though (?) since most of parts of the tutorial are pre-yadism. Maybe we can just port the part in which we compare with data? So there would be two main sections:
- the current tutorial -> Getting started
- comparison to measurements

Or do you want to really start from the basics->yadism?

felixhekhorn:
> Maybe we can just port the part in which we compare with data?

yes, that should be sufficient

Radonirinaunimi:
> > Maybe we can just port the part in which we compare with data?
> 
> yes, that should be sufficient

Ok, I will add that part here.

giacomomagni:
To remember some TODOs left in the documentation: 

* [ ] TMC, both  for theory section and implementations
* [ ] Describe `kernels`  and `ESFResult`
* [ ] Add something about FFN0 in the theory section and clarify that FONLL is not implemented only by Yadism

PR #232, opened 2023-10-12
Observable card generators for pDIS

Radonirinaunimi:
@Toonhasenack, please implement and update here the implementation of the observable cards generator. 

PS: Recall that to generate the observable cards you need to be in `extra/data` and run `python -m machinery generate E155/*`.

The full list of datasets can be found in https://github.com/NNPDF/nnpdf/pull/1816

PR #266, opened 2024-03-11
Add EIC projection measurements for polarised fits

Radonirinaunimi:
@toonhasenack Here you find the kinematics for the projected measurements of various observables at the EIC. For each of these measurements, you have to implement the observable generator. There are two main classes of measurements:

**EIC and EIcC:**

These are projection measurements for $A_1^c \equiv g_1^c/F_1^c$ on electron-proton scattering. That is, the intermediate observables that we need to compute from Yadism are now `g1_charm` and `F1_charm`. In terms of computations, the pipeline (in your `pineverse`) is exactly the same as `G1F1RATIO`.  In terms of dataset naming, we could call them `<EIC/EIcC>_NC_<SQRTS>GEV_EP_G1F1CHARMRATIO` where is again what you put in `observable_name: G1F1CHARMRATIO` later on in the data implementation.

The naming convention of the `.txt` files are as follow: 

`<EIC/EIcC>_<electron_beam_energy>_<proton_beam_energy>_A1C_<luminosity>fb-1.txt`

while each column represent `x, Q2, Asbolute_uncertainty`.

**Athena**

These are projection measurements for $A_{LL} \equiv D(y) g_1^c/F_1^c$ on electron-proton scattering where 

$$D=\frac{y(2-y)}{y^2 + 2(1-y)}$$

which depends on the inelasticity `y` is the depolarization factor. The computation of this observable is slightly more involved that what we've had to computed before. First, we need to compute the intermediate observable as before using `g1` and `F1`, then compute the ratio between $g_1/F_1$ also as before by applying them as k-factors. But now we also need to compute `cfactor` normalization by computing instead $D(y)$ rather than $1/(2x)$.

The corresponding datasets could be named: `ATHENA_NC_<SQRTS>GEV_EP_ALL` where again `ALL` is what you put in `observable_name: ALL` later on in the new commondata implementation. For this pseudo-measurement, the name of each column is given in the headers.

**Important:**  There is only one single `.xlsx` file iven that there are different beam energies with overlapping $(x, Q^2)$ bins you need to separate them in terms of the center-of-mass energy - hence why the target grids are separated as [follow](https://github.com/NNPDF/yadism/blob/EIC_pDIS/extras/data/ATHENA/metadata.yaml).

PS: As you can see, the naming of the experiments are not fully consistent (given that Athena would be part of ePIC which is again part of EIC), if you have better name please suggest.


alecandido:
> **Important:** There is only one single `.xlsx` file iven that there are different beam energies with overlapping

Converting it is not option? It could be CSV, JSON, YAML, or even `.npy`...

extras/data/ATHENA/metadata.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+# For Sqrt_s=29 GeV and beam energies (5, 41) GeV
+ATHENA_NC_29GEV_EP_G1:
+  description: polarized structure function $g_1$ from ATHENA projections
+  y_label: 2xg1
+  y_label_tex: $2xg_1$
+  hepdata: null
+
+ATHENA_NC_29GEV_EP_F1:
+  description: polarized structure function $F_1$ from ATHENA projections
```
felixhekhorn, line 12:
```suggestion
  description: unpolarized structure function $F_1$ from ATHENA projections
```
you didn't copy from the latest version of #232 :upside_down_face:

felixhekhorn:
you don't want to base this on top of #232 ?

extras/data/ATHENA/metadata.yaml
Radonirinaunimi: Thanks! I did only copy internally the description for `g1` 😅 Fixed now in [99839ff](https://github.com/NNPDF/yadism/pull/266/commits/99839ff3701b1b69025c18d60b5468af4cef42da).

Radonirinaunimi:
> you don't want to base this on top of #232 ?

I was not sure actually - the only tiny reason in basing it on top of master is that both don't have to be merged in order. But if you like it more on top of #232 I can change it.

extras/data/ATHENA/metadata.yaml
giacomomagni: At some point we will converge here 😬

felixhekhorn:
> > you don't want to base this on top of #232 ?
> 
> I was not sure actually - the only tiny reason in basing it on top of master is that both don't have to be merged in order. But if you like it more on top of #232 I can change it.

well, there you said your waiting for NNPDF, so surely here as well - and you may want to use `dump_polarized` also here ...

Radonirinaunimi:
> > > you don't want to base this on top of #232 ?
> > 
> > 
> > I was not sure actually - the only tiny reason in basing it on top of master is that both don't have to be merged in order. But if you like it more on top of #232 I can change it.
> 
> well, there you said your waiting for NNPDF, so surely here as well - and you may want to use `dump_polarized` also here ...

Branching off of #232 done. It was a bit dirty but that's now changed.

extras/data/machinery/generate/ATHENA.py
```diff
@@ -0,0 +1,49 @@
+from .utils import *
+import numpy as np
+
+
+def dump(src_path, target):
```
felixhekhorn, line 5: Please merge the three `dump` (this one and the other two) with `dump_polarized` by adding argument and/or abstract workflow

toonhasenack: Hi Felix, could you clarify why? The reason that ATHENA, EIC and EIcC are not in dump_polarized is that their cases are completely custom, as the data is read from Yadism instead of commondata.

felixhekhorn: mh? this has nothing to do with "data"* (in the NNPDF sense) - what you're supposed to do here is to prepare the yadism observable cards; for what you are using these is beyond `yadism` scope (i.e. if you use it just for the theory predictions or also for the pseudo data generatation). And just by looking at the code I can see that there is very strong overlap

PS: *fine calling this folder `data` is a bit misleading, maybe `real_life_experiments` would be better :upside_down_face:

toonhasenack: Okay let me clarify what I mean by data: the format of the projection datafiles inside Yadism is completely different from the other polarised sets (observe already the difference between .txt and .yaml). This is so, because all polarised (non-projective) data inside Yadism is directly computed from commondata and hence in similar format. If you really see the advantage of adding three distinct cases to the dump_polarized, I don't mind doing it, but I don't see the advantage myself to be honest.

felixhekhorn: Look:
```
$ diff <(sed -n "68,104p;105q" utils.py) <(sed -n "6,40p;41q" ATHENA.py)
1,2c1,2
< def dump_polarized(src_path, target):
<     """Generate the input card for polarized measurements.
---
> def dump(src_path, target):
>     """Generate the input card for ATHENA measurements.
18,21c18,19
<     data = load(src_path, 0, ["x", "Q2"])
<     dict_kins = [
<         dict(x=d["x"]["mid"], y=d["y"]["mid"], Q2=d["Q2"]["mid"]) for d in data
<     ]
---
>     data = np.genfromtxt(src_path)
>     dict_kins = [dict(x=float(d[0]), y=float(d[2]), Q2=float(d[1])) for d in data]
```
the first two lines are completely irrelevant and for the second two I can see some trivial mapping. All the other 13 lines in that function are identical copies. Don't repeat yourself.

extras/data/machinery/generate/utils.py
```diff
@@ -81,14 +81,39 @@ def dump_polarized(src_path, target):
...
+        observable_name = "F1_charm" if "_F1" in target.parent.name else "g1_charm"
+
+    else:
+        data = load(src_path, 0, ["x", "Q2"])
+        dict_kins = [
+            dict(x=d["x"]["mid"], y=d["y"]["mid"], Q2=d["Q2"]["mid"]) for d in data
+        ]
+
+        obs["PolarizationDIS"] = 0.0 if "_F1" in target.parent.name else 1.0
+        observable_name = "F1_total" if "_F1" in target.parent.name else "g1_total"
```
felixhekhorn, line 115: nice try, but no! okok, I admit it - this is already better than the last time, but not good enough yet :upside_down_face: there is still a lot of repetition and together we can get rid of it! Let's start from the end:
- 4th line: how about taking `heavyness` as argument to the function? default it to `"total"` and you save 75-5 character of repetition
- 3rd line: 69 character of repetition
- 1st and 2nd line: fine we need to think a tiny bit more - but we are developers, we can do it! :upside_down_face: e.g. we can simply harmonize the file format: i.e. parse the EIC kinematics file and make them compatible with the others. Another possibility would be to split the function and eventually take `dict_kins` as input, or many possibilities more - the Italians say at this point: "Tutte le strade portano a Roma" (which you may recognize from Asterix)

Don't repeat yourself.

felixhekhorn (changes requested):

extras/data/ATHENA/athena_105gev_ep
felixhekhorn, line 1: Please add a file extension here and to the others

extras/data/machinery/generate/utils.py
```diff
@@ -81,15 +81,34 @@ def dump_polarized(src_path, target):
...
+
+    else:
+        data = load(src_path, 0, ["x", "Q2"])
+        dict_kins = [
+            dict(x=d["x"]["mid"], y=d["y"]["mid"], Q2=d["Q2"]["mid"]) for d in data
+        ]
+
+    heavyness = "total"
+    if "EIC" in target.parent.name or "EIcC" in target.parent.name:
+        heavyness = "charm"
```
felixhekhorn, line 104: again, better but not good enough yet :upside_down_face: we can save these hard coded lines ... 1) we should always aim to avoid hardcoded stuff, as this will hinder the extensibility greatly 2) having this `if` here make the EIC implementation not self-contained, i.e. part of the logic is sitting here;

let me repeat what I said: "let's make `heavyness` an argument`; this means
- change the signature of this function from `dump_polarized(src_path, target)` to `dump_polarized(src_path, target, heavyness="total")`
- and then in the driver file you replace `from blub.bla import dump` by `from blub.bla import dump as raw_dump [...] def dump(src_path, arget): return raw_dump(src, target, "charm")`

extras/data/machinery/generate/utils.py
```diff
@@ -81,15 +81,34 @@ def dump_polarized(src_path, target):
...
-    dict_kins = [
-        dict(x=d["x"]["mid"], y=d["y"]["mid"], Q2=d["Q2"]["mid"]) for d in data
-    ]
+    data = np.genfromtxt(src_path)
+
+    if (
+        "ATHENA" in target.parent.name
+        or "EIC" in target.parent.name
+        or "EIcC" in target.parent.name
+    ):
```
felixhekhorn, line 93: same holds as for my comment on `heavyness` - let's drop these ifs to a "higher level", i.e. outside of this function into the caller; see my previous comment about the options (and the freedom you have) - if you insist on a switch here, you could do it by file type (but NOT by name) and don't change the default how data is read (because the new way is actually the minority)

PR #270, opened 2024-03-22
Generate kfactors for polarized observables

felixhekhorn:
Continuation of https://github.com/NNPDF/pineko/pull/132

This is not my PR, but it should have been opened by @toonhasenack .


Radonirinaunimi:
Since now `n3fit` and `vp` can deal with different initial states we no longer need this.

PR #272, opened 2024-03-28
Update adani

niclaurenti:
Updating `adani`.
Now the needed functions are accessed through the abjects `adani.ApproximateCoefficientFunction` and `adani.HighScaleSplitLogs`.

@giacomomagni regarding the new approximation for $a_{Qg}^{(3,0)}$, do we want to use it as a default or we want the old one (I guess that the aswer is the new one but better asking)?

So far I didn't push the new version on the PyPi so this PR cannot be merged (and the tests are going to fail)


giacomomagni:
> @giacomomagni regarding the new approximation for aQg(3,0), do we want to use it as a default or we want the old one (I guess that the aswer is the new one but better asking)?
> 
we do want to use the new one!! It has a much better understanding of the small-x regions, see https://arxiv.org/pdf/2403.00513.pdf fig 2

extras/heavy_em_n3lo/yad_grids.py
```diff
@@ -11,15 +11,35 @@
...
+    raise ValueError("Set channel to 'g' or 'q'")
+
+order = int(sys.argv[5])
+
+hs_version = sys.argv[6]
+if hs_version not in ["klmv", "abmp", "abbdvss"]:
+    raise ValueError("Set hs_version to 'abmp' or 'abbdvss'")
+# klmv = Kawamura, Lo Presti, Moch, Vogt: approximation from [arXiv:1205.5727]
+# abmp = Alekhin, Blumlein, Moch, Placakyte: approximation from [arXiv:1701.05838]
+# abbdvss = Ablinger, Behring, Blumlein, De Freitas, von Manteuffel, Schneider, Schonwald: approximation from [arXiv:2403.00513]
```
giacomomagni, line 29: The approximation is actually mine, based on their result, but not exactly them... 
(I mention this before someone will complain, not because I care of the authorship)

niclaurenti: ok sure. tomorrow I will fix both in yadism and adani

src/yadism/coefficient_functions/asy/f2_nc.py
```diff
@@ -7,6 +7,8 @@
 
 
 class AsyLLGluon(pc.NeutralCurrentBaseAsy):
+    hs3 = adani.HighScaleSplitLogs(3, '2', 'g', "abmp")
```
giacomomagni, line 10: not sure if here it's nicer to have this the top level, 
creating an additional class and avoid repetition 

```
class AsyGluon(pc.NeutralCurrentBaseAsy):
        hs3 = adani.HighScaleSplitLogs(3, '2', 'g', "abmp")
...
class AsyLLGluon(AsyGluon):

....
```

niclaurenti: Ok done!

giacomomagni (approved):

PR #274, opened 2024-04-06
Fix leftovers of #272

niclaurenti:
Fixing some leftovers of PR #272: there were some bugs in the file `yad_grids.py` and in the N3LO asymptotic limits from `adani`.
Moreover I updated the `adani` version since I didn't push the 'gm' approximation for the high scale limit


src/yadism/coefficient_functions/asy/fl_nc.py
```diff
@@ -35,7 +35,9 @@ def cg_NLL_NNLO(z, args):
     def N3LO(self):
         def cg_NLL_N3LO(z, args):
             L = -args[0]
-            return self.hs3.NLL(z) * L**2
+            # this term does not depend on nf so setting it to zero
+            nf = 0
+            return self.hs3.NLL(z, nf) * L**2
```
giacomomagni, line 40:
```suggestion
            return self.hs3.NLL(z, nf=0) * L**2
```
I'd add it in inline.

niclaurenti: Unfortunately in the pywrap.cc I didn't add `py::arg("nf")` for that function so it cannot be called with `nf=0`.
I will add it so in a future release it will be possible

niclaurenti: https://github.com/niclaurenti/adani/blob/ce7a7ec745c2643929f4b9e7aeaa66058c179b65/pywrap/pywrap.cc#L286

giacomomagni (approved):

Issue #276, opened 2024-04-12
Massless NC N3LO flavor decomposition is wrong


giacomomagni:
The N3LO factorization of the NC ($ZZ$ contributions) coupling is wrong. Let me explain the problem.
At N3LO (for the first time) there are 2 different flavor classes of diagrams (or topologies) as reported in Fig1 [1].

* The `fl2` diagrams, in which the incoming and outgoing boson is coupled to the **same** fermionic line (these diagrams are the only one present up to NNLO)
* The `fl11` diagrams, in which  the incoming and outgoing boson is coupled to **different** fermionic lines

1) For CC the second class of diagrams vanishes as a single $W$ can't couple to flavor loop, so we are fine. 
2) For NC PV we are also fine as this class does not shows up (or better, if it shows up (is allowed by symmetry ?!) 
it has not been computed yet as we are currently getting the F3 NC coefficient from the CC $\nu - \bar{\nu}$. ). 

## `fl11` issue 

Now all our references deals only with the photon exchange case, in which only `VV` couplings are present, 
and here we are making our mistake. Following [2], see eq 55 and discussion above, the coefficient functions are factorized as follow:

$$g_{fl2} ( C_{fl2} + g_{fl11} / g_{fl2} C_{fl11} ) \quad \text{Eq (1)}$$

where $g_{fl2}$ is the coupling of the flavor class `fl2` i.e. the standard one, while $g_{fl11}$ is the coupling of the new flavor which are then given in table 2 of [2] for photon exchange as a ration. However, in the case of Z exchange both Vector-vector and Axial-vector type coupling are present.

Currently we factorize the partonic coupling as

$$F_{i}^{bb'} = (g^b_A g^{b'}_A + g^b_V g^{b'}_V) C_i \otimes f_i \quad i = g,q$$

so we have two different solutions. 

* Either we extend this equation to be:

```
F_{i}^{bb'} = [ 
               (g^b_{A,fl2} g^{b'}_{A,fl2} + g^b_{V,fl2} g^{b'}_{V,fl2}) C_{i,fl2} +
               (g^b_{A,fl11} g^{b'}_{A,fl11} + g^b_{V,fl11} g^{b'}_{V,fl11})  C_{i,fl11} 
        ] \otimes f_i \quad i = g, q
```

* Either we recover Eq(1) and we compute all the necessary $g_{fl11} / g_{fl2}$ ratios which can be quite complicated 
especially for the ZZ channel where both AA and VV contribute.

[...]

PR #280, opened 2024-04-18
N3LO flavor decomposition

giacomomagni:
This PR is to address the issue #276 regarding `fl02` diagrams, introducing the Valence channel 
and fixing the coupling of the `fl11` flavor class.


src/yadism/coefficient_functions/light/partonic_channel.py
```diff
@@ -4,5 +4,6 @@
 class LightBase(pc.PartonicChannel):
     """Light partonic channel base class."""
 
-    def __init__(self, ESF, nf):
+    def __init__(self, ESF, nf, is_fl11=False):
         super().__init__(ESF, nf)
+        self.is_fl11 = is_fl11
```
felixhekhorn, line 9: just a minor detail: this name makes me suspicious ... what I would do instead is define a new partonic channel (i.e. this _class is_ fl11), which only contributes at N3LO - so instead of the `if` branching, just make new contribution (with plain return); this business should be really additive and so we should treat it like this

src/yadism/coefficient_functions/light/n3lo/xclsg3p.py
```diff
@@ -39,7 +39,7 @@ def cls3a(y, args):
         + 9.773 * dl
         + y * dl * (363.8 + 68.32 * dl)
     ) * y
-    # Note here the source file cointain a typo and the
+    # Note here the source file contain a typo and the
```
felixhekhorn, line 42:
```suggestion
    # Note here the source file contains a typo and the
```

src/yadism/coefficient_functions/light/n3lo/xc2ns3p.py
```diff
@@ -63,23 +62,28 @@ def c2np3a(y, args):
...
         )
-        + fl11
-        * nf
-        * (
+    )
+    return res
+
+@nb.njit("f8(f8,f8[:])", cache=True)
+def c2np3a_fl11(y, args):
+    """The math:`f_{l11}`: regular piece of the non singlet coefficient."""
```
felixhekhorn, line 70: it is $fl_{11}$, right?
```suggestion
    """The math:`fl_{11}`: regular piece of the non singlet coefficient."""
```
(similar for others)

src/yadism/coefficient_functions/light/f2_nc.py
```diff
@@ -1,3 +1,4 @@
+from .. import partonic_channel as epc
```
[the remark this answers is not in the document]
giacomomagni: I actually think we can avoid adding all this empty valence channels and this will never shows up in f2. 
Do you agree? 
Same for gluon and singlet in PV structure functions.

src/yadism/coefficient_functions/light/partonic_channel.py
giacomomagni: This is ofc possible. 
In the current implementation I tried to use `min_order` to silence the previous orders, 
but I can create new partonic channels if this is clearer.

felixhekhorn: I would say a single class (e.g. `GluonFL11`) is better than `min_order` + `if` - at least to me; and you may be able to reuse it this way ...

src/yadism/coefficient_functions/light/f2_nc.py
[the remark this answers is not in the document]
felixhekhorn: you can change if `kernels` does not access such a class (which it typically does in a observable independent matter i.e. `f2_nc` is dynamic)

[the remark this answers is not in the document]
giacomomagni: I disagree, now that we are complicating the structure and we have `is_pv` around, be sure not ending up in `f3.Gluon` would be an nice check.

src/yadism/coefficient_functions/light/partonic_channel.py
giacomomagni: okay fine

giacomomagni:
@felixhekhorn the implementation should be ready and working. 
I'm not sure why workflows are not running anymore.

docs/source/theory/nonsinglet.rst
```diff
@@ -111,20 +100,93 @@ Consider the following hypothesis on the number of flavors:
...
+-------------
+
+The decomposition of the quark sector in different partonic channels
+has the advantage to facilitate the relations with higher orders
+QCD corrections.
+
+- :math:`C_{ns}` is always the leading contribution as correspond
+  to diagram in which the incoming flavor is coupling directly to the
+  electroweak boson.
+- :math:`C_{g}` is  suppressed suppressed by :math:`\mathcal{O}(a_s)`
```
felixhekhorn, line 178:
```suggestion
- :math:`C_{g}` is suppressed by :math:`\mathcal{O}(a_s)`
```

docs/source/theory/nonsinglet.rst
```diff
@@ -111,20 +100,93 @@ Consider the following hypothesis on the number of flavors:
...
+  as the gluon need to radiate a quark-antiquark pair before coupling
+  with a electroweak boson.
+- :math:`C_{ps},C_{v}` are suppressed suppressed by :math:`\mathcal{O}(a_s^2)` or
```
felixhekhorn, line 181:
```suggestion
- :math:`C_{ps},C_{v}` are suppressed by :math:`\mathcal{O}(a_s^2)` or
```

docs/source/theory/nonsinglet.rst
```diff
@@ -111,20 +100,93 @@ Consider the following hypothesis on the number of flavors:
...
+  :math:`\mathcal{O}(a_s^3)` respectively as they are related to diagrams
+  where the incoming flavor line is not coupling directly with the electroweak boson.
+
+
+From N3LO an new class of diagrams, called :math:`fl_{11}`, can appear for
+the parity conserving structure functions, bot in the quark and gluon sector
```
felixhekhorn, line 187:
```suggestion
From N3LO on a new class of diagrams, called :math:`fl_{11}`, can appear for
the parity conserving structure functions, both in the quark and gluon sector
```

docs/source/theory/nonsinglet.rst
```diff
@@ -111,20 +100,93 @@ Consider the following hypothesis on the number of flavors:
...
+:cite:`Larin:1996wd`. In these diagram the incoming and outgoing bosons are
```
felixhekhorn, line 188:
```suggestion
:cite:`Larin:1996wd`. In these diagrams the incoming and outgoing bosons are
```

src/yadism/coefficient_functions/light/kernels.py
```diff
@@ -197,15 +193,14 @@ def nc_fl11_weights(coupling_constants, Q2, nf, skip_heavylight=False):
...
-    # and the contribution of the singlet PDF to the S coefficient
-    # do not cancel completely.
+    # compute gluon
     ch_av = tot_ch_sq / len(pids)
-    s_partons = {q: ch_av - ns_partons[q] for q in [*pids, *(-q for q in pids)]}
-    return {"ns": ns_partons, "g": {21: ch_av}, "s": s_partons}
+
+    # NOTE: since here we are implementing the plain coupling
+    # Q1 * Tr[Q2] / nf, we have not th quark sector into singlet
+    # and non singlet
```
felixhekhorn, line 205: mh? this sentence doesn't make sense to me ... it seems to miss a verb in the second half ...

giacomomagni: done in https://github.com/NNPDF/yadism/pull/280/commits/2614f0282e40985eeba9a2676609114786bb1707

src/yadism/coefficient_functions/light/n3lo/xc3ns3p.py
```diff
@@ -69,7 +64,19 @@ def c3nm3a(y, args):
             - dl * dl1 * (39.99 + 5.103 * dl - 16.30 * dl1)
             + 0.0647 * y * dl**4
         )
-        + fl02
+    )
+    return res
+
+
+@nb.njit("f8(f8,f8[:])", cache=True)
+def c3nsv3a(y, args):
```
felixhekhorn, line 72:
```suggestion
def c3nsv3a(y, args):
    # contributions from flavor class fl02
```
let's be explicit (I know it is written below in code, but we could even drop that to stress this is the associated coefficient)

felixhekhorn (approved):
- take a look to my changes in 23bd478e
- maybe we should add some more tests to `test_partonic_coupling_fl11` after all this mess?

else looks good to me

felixhekhorn:
Re workflows:
```
Invalid workflow file: .github/workflows/unittests.yml#L12
error parsing called workflow
".github/workflows/unittests.yml"
-> "N3PDF/workflows/.github/workflows/python-poetry-tests.yml@v2"
```
we need to check there ...

giacomomagni:
> * maybe we should add some more tests to `test_partonic_coupling_fl11` after all this mess?
> 
Yes indeed I'd like to add a test that check that our implementation is equivalent to the sum of singlet and non-singlet with the weights given in literature, for photon exchange, but I didn't had time yet.

felixhekhorn:
> Re workflows:
> 
> ```
> Invalid workflow file: .github/workflows/unittests.yml#L12
> error parsing called workflow
> ".github/workflows/unittests.yml"
> -> "N3PDF/workflows/.github/workflows/python-poetry-tests.yml@v2"
> ```
> 
> we need to check there ...

I attempted a fix in eae37bca and they seem to be back ... I don't know why we see this change now, somehow the redirect expired I guess? We will have to propagate this fix around to the others (e.g. [eko](https://github.com/NNPDF/eko/actions/runs/8892211170))

giacomomagni:
> * maybe we should add some more tests to `test_partonic_coupling_fl11` after all this mess?

done in [f43d8b5](https://github.com/NNPDF/yadism/pull/280/commits/f43d8b55e63792afbfe5eb83c8f20ce742225eba), 
I don't know how to test it further for NC as the leptonic side will also mess up...

giacomomagni:
> Add something for the bugs you found in the last commits!

I've added a test, in  https://github.com/NNPDF/yadism/pull/280/commits/c6f78bbba8efb5bca29051289b639ead8a493312 could you please check it makes sense?? There are quite a few -1 sign in CC.

I'm still working on the other test...

felixhekhorn:
> I've added a test, in [34fbe0a](https://github.com/NNPDF/yadism/pull/280/commits/34fbe0a7477ffd8fa0930285ff0687f0c86abdd2) could you please check it makes sense?? There are quite a few -1 sign in CC.
> 
> I'm still working on the other test...

okay, so what you are saying is that in fl02 the "boson loop" is decoupled from the PDF and hence the coefficient function can not depend on the boson type (i.e. $W^\pm$), right? Sounds correct to me ...

giacomomagni:
I'll merge as soon as tests are finished.

PR #294, opened 2024-07-05
Different hotfixes

giacomomagni:
This PR includes 3 different hotfixes to make yadism running again. 

* Fix new pineappl convention.
* Add empty IC coefficient functions for polarized structure functions (needed since we have dropped `IC`).
* Update theory cards compatibility layer, according to the new NNPPDF runcards. 

The only change that is not strictly necessary is setting `RenScaleVar,FactScaleVar` to `True` if the key is not present. 
I find this change quite handy to speed up the computation of scale varied theories, but if you don't like it
I'm happy to drop it.


Radonirinaunimi (approved):
LGTM! I don't have a preference re the `_ScaleVar` part as I am not myself extensively generating SVed theories.

felixhekhorn (approved):

PR #303, opened 2024-10-04
add explanation for iron Z and A values

RoyStegeman:


RoyStegeman:
Wait, maybe I'm wrong since I don't find the values in the paper: https://arxiv.org/pdf/hep-ex/0102049

alecandido:
> Wait, maybe I'm wrong since I don't find the values in the paper: [arxiv.org/pdf/hep-ex/0102049](https://arxiv.org/pdf/hep-ex/0102049)

As said in private communication, those numbers definitely came from APFEL, and the "I don't know comment" was the conservative one.
Maybe it could be just rephrased in a softer way... or left as it is...

RoyStegeman:
Hi @enocera, do you perhaps know the source of the A and Z values for iron used in APFEL (and now yadism) predictions of NuTeV? I tried goodling and searching for those values in the nutev papers, but could not find their origin.

enocera:
Hi @RoyStegeman the numbers are taken from A. Mason's PhD thesis, available at this link
https://inspirehep.net/files/15b639af43688d59f1ba65ec9cde7ec7
see in particular page 323 there.

RoyStegeman:
Thanks, I would not have found that in a million years!

alecandido:
> Thanks, I would not have found that in a million years!

Me too, thanks @enocera

alecandido (approved):

PR #308, opened 2024-11-01
extend the n3lo heavy grids up to xi=1e9

RoyStegeman:
For the fiatlux grids we need to compute grids up to $Q^2=10^8$ (see [pincards](https://github.com/NNPDF/pinecards/tree/master/FIATLUX_DIS_F2)) while currently they have been computed up to $\xi=Q^2/m_c^2=10^4$. 

Here I extend the grid up to $\xi=10^9$ while keeping roughly the same point density because 49*12/7=84. 

I believe @niclaurenti did something similar, keeping the grids only locally, but I don't see a reason not to just extend them.


giacomomagni (approved):
LGTM, thanks. 
If this works you can merge for me!

felixhekhorn:
I guess there is nothing wrong with this - except that it seems we are bound by the NNLO precision instead ...

RoyStegeman:
Used in the computation of the photon at alphas=0.119 compared to the one done by Niccolo at alphas=0.118: https://vp.nnpdf.science/YFzb3pnfQD-c6IMhyRuMAg==/

PR #309, opened 2024-11-07
Fix CKM as list

giacomomagni:
In the new nnpdf runcards `CKM` is a list, however it was not passed correctly to yadism. 

I think here we have been lucky no one was doing CC DIS grids since a while...

### DIFF
--- a/src/yadism/coefficient_functions/coupling_constants.py
+++ b/src/yadism/coefficient_functions/coupling_constants.py
@@ -415,6 +415,8 @@ class CouplingConstants:
         ckm_matrix = theory["CKM"]
         if isinstance(ckm_matrix, str):
             ckm_matrix = CKM2Matrix.from_str(ckm_matrix)
+        elif isinstance(ckm_matrix, list):
+            ckm_matrix = CKM2Matrix(np.array([ckm_matrix]) ** 2)
         theory_config = {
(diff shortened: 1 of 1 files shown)


RoyStegeman (approved):
Thanks!

src/yadism/coefficient_functions/coupling_constants.py
```diff
@@ -477,6 +479,10 @@ def __repr__(self):
         """Return string representation."""
         return "CKM(" + str(self.m).replace("\n", "") + ")"
 
+    def __eq__(self, other) -> bool:
+        """Equal method."""
+        return (self.m == other.m).all()
```
felixhekhorn, line 484: @giacomomagni that's not good (float and `==`) - it should rather be `np.assert_all_close`

giacomomagni: sure feel free to adjust it, I added this mostly to make a test not because we really need it

RoyStegeman: I think just `np.allclose`, but you're right. Then again, the input is always the exact same and we don't do any operations on it (or the same on all, namely taking the square only once), so comparing two should really always give the same result even accounting for finite precision due to floats.

PR #310, opened 2024-11-14
replace NaNs in output with 0

RoyStegeman:
To produce the fiatulux grids we use points at small-x (x=1e-6), where LeProHQ may return NaNs. @niclaurenti used a [separate branch](https://github.com/NNPDF/yadism/tree/hack_nans) to produce these grids, but I believe we should be able to use master going forward so I implement the function here.


felixhekhorn:
The moment you fix black, I think this is good to go

felixhekhorn:
after merging, I guess we should make a tag to make @scarlehoff happy - see https://github.com/NNPDF/pinefarm/pull/87

it will be a minor - see https://github.com/NNPDF/yadism/compare/v0.13.4...master

PR #346, opened 2026-02-02
Fix NaN appearing in heavy N3LO grids

Radonirinaunimi:
The following exposes the issues that plague N3LO heavy grids with `NaN`. 9e4f55ac1f3664cae712c27c8d1478043e2639b7 provides a simple fix but might not be exactly what we wnat. There were mainly two problems working together.

**Problem 1: NaN values in N3LO heavy quark grid files**

The pre-computed N3LO coefficient function grids in [src/yadism/coefficient_functions/heavy/n3lo/grids/](https://github.com/NNPDF/yadism/tree/master/src/yadism/coefficient_functions/heavy/n3lo/grids) contain `NaN` values. The problem is that when `RectBivariateSpline` [heavy/n3lo/\_\_init__.py](https://github.com/NNPDF/yadism/blob/master/src/yadism/coefficient_functions/heavy/n3lo/__init__.py) is constructed from a grid containing `NaN` values, the spline fitting algorithm could propagate the `NaN` values through the entire set of spline coefficients.

$\Longrightarrow$ Interpolate/Extrapolate the small regions that contain `NaN` using nearest neighbor fit.

**Problem 2: `replace_nans_with_0` never run on generic/compound observables**

The safety net (`replace_nans_with_0`) which was meant to safeguard this in the first place never gets called for generic observables (which in the case for NMC is `F2_total`).

$\Longrightarrow$ A simple fix for this is to check the actual observable:
```python
  if not observable_name.ObservableName.is_valid(observable):
      continue
```

### DIFF
--- a/src/yadism/coefficient_functions/heavy/n3lo/__init__.py
+++ b/src/yadism/coefficient_functions/heavy/n3lo/__init__.py
@@ -35,7 +35,7 @@ def interpolator(coeff, nf, variation):
 
     # load grid
     coeff = np.load(grid_path / grid_name)
-    if np.isnan(coeff).sum() != 0:
+    if np.isnan(coeff).any():
         coeff = fill_nans_nearest_neighbor(xi_grid, eta_grid, coeff)
     grid_interpolator = RectBivariateSpline(xi_grid, eta_grid, coeff)
 


felixhekhorn:
> comparing

ehm this is hard to read - can't you please do `pineappl diff`?

PR #349, opened 2026-02-10
Drop asv

felixhekhorn:
It hasn't been used since an initial attempt and makes problems in #348 - let's drop it all together


felixhekhorn:
For the record: #348 failed due to https://github.com/pypa/setuptools/issues/5174

PR #374, opened 2026-06-08
Fix bug and mention lhapdf installation in tutorials

eljaalha:
- Update syntax in tutorial to match PineAPPL v1
- mention LHAPDF is required to get actual numbers


Radonirinaunimi (changes requested):
Thank you very much @eljaalha for updating the tutorials. While you are at it, could you please also address the following?

- Add a comment that the following also achieves the same thing as defining manually the types of convolutions:

```python
yadism_data = grid.convolve(grid.convolutions, [pdf.xfxQ2], pdf.alphasQ2)
```

- Add `lhapdf.setVerbosity(0)` to suppress the LHAPDF banner

Otherwise, the rest LGTM.

felixhekhorn:
Thanks for jumping in @Radonirinaunimi I was about to suggest you as a reviewer :innocent:

eljaalha:
I made the changes suggested by @Radonirinaunimi. I figured there isn't any reason to define the convolution types manually in the compare_data tutorial so I just used grid.convolutions there. I kept the manual definition in getting_started and added the more concise way of doing it as a comment there as suggested, so the reader knows that both ways are possible.

Radonirinaunimi (approved):
This is much better, yes. Thanks!

Issue #260, opened 2024-01-30
Drop intrinsic configuration and activate it always


felixhekhorn:
IC can be always active and the decision of whether to use it or not left to the user upon PDF convolution;

I think the only affected line (apart from passing the config around) is this one:
https://github.com/NNPDF/yadism/blob/c06562d6434672534801a6cd06687499660ff0ec/src/yadism/coefficient_functions/__init__.py#L134-L145

which we de-indent and finish.

PR #382, opened 2026-06-22
Suppress intrinsic bottom and top contributions

Radonirinaunimi:
This is an alternative approach to https://github.com/NNPDF/pineko/pull/253 to address the issue discussed in #380 by suppressing the intrinsic bottom/top contributions at the level of the grids.


felixhekhorn:
To be explicit: this would revert #262 #260

PR #384, opened 2026-06-30
Reproduce 4.0 aN3LO methodology for CF computations

Radonirinaunimi:
@kamillaurent, @jekoorn,  This is the branch that could be used to compute the aN3LO DIS coefficient functions that reproduce the 4.0 predictions. The idea would be to recompute the theories with 4.1 settings but using the 4.0 methodology for the N3LO approximation.


Radonirinaunimi:
But perhaps before recomputing the theories, let's wait for the Code Meeting this afternoon for the proper action plan.

jekoorn:
thanks! I agree re code meeting.

PR #383, opened 2026-06-24
update to python 3.14, drop 3.10, require eko 0.15.4

jekoorn:
* update to python 3.14
* drop python 3.10, 
* require eko 0.15.4

### TODO

- [x] Please fix .readthedocs.yaml (as in eko)
- [x] import https://github.com/NNPDF/yadism/pull/381 https://github.com/NNPDF/yadism/pull/379
- [x] add Changelog entry
- [x] make unit test pass

Issue #162, opened 2022-12-02
Benchmark Workflow


AleCandido:
Ok, blaming in Git it seems like that the problem described in https://github.com/NNPDF/yadism/pull/161#issuecomment-1335716316 it is my fault, but something I did one year ago.

This means that `yadism` is approximately unmaintained, and almost no development happened in the meanwhile...

However, we need to drop asserts, since we never developed a suitable benchmark metric to automatically applied, see #83.
For this, I propose:
- [ ] add in `banana` a no-assert mode, to me triggered by the workflow
- [ ] add in `banana` report extraction, such that we can upload them as artifacts in the CI
  - the other, sensible, option is to upload the whole database, but I'd rather do both, since it is less clear to people how to do both...

Even though once you have the db and the extraction method this should be trivial (if well-documented), the workflow outcome should be intuitive to consume on its own.
It is as much trivial to run the conversion in the CI and upload both.

felixhekhorn:
actually, should we transfer this to `banana`? (which has even less traffic by the way, but this means it is working just fine, I'd say ...)

AleCandido:
The issue at the moment is here, so I'd keep where it arose.

You can open a further issue in `banana`, exporting the todo items from here.

PR #388, opened 2026-08-12
Decorrelate N3LO IHOU

### LINKED ISSUE #328: N3LO IHOU are correlated
`n3lo_cf_variation` seems to be a single number, although I believe it should be 4 numbers: `(C2g, C2q, CLg, CLq)`, i.e. one per coefficient we want to vary - similar to the case in eko

jekoorn:
Closes #328 

n3lo_cf_variation seems to be a single number, although I believe it should be 4 numbers: (C2g, C2q, CLg, CLq), i.e. one per coefficient we want to vary

### DIFF
--- a/src/yadism/coefficient_functions/asy/kernels.py
+++ b/src/yadism/coefficient_functions/asy/kernels.py
@@ -109,6 +109,10 @@ def generate_heavy_asy(esf, nf, pto_evol, ihq):
         if not is_pv:
             n3lo_cf_variation = esf.info.theory["n3lo_cf_variation"]
             for c, channel in (("g", "Gluon"), ("s", "Singlet")):
+                if c == "g":
+                    variation = n3lo_cf_variation[0]
+                if c == "s":
+                    variation = n3lo_cf_variation[1]
                 for res in range(pto_evol + 1):
                     name = "Asy" + ("N" * res) + "LL" + channel
                     for av in ("AA", "VV"):
@@ -119,7 +123,7 @@ def generate_heavy_asy(esf, nf, pto_evol, ihq):
                                     esf,
                                     nf,
                                     m2hq=m2hq,
-                                    n3lo_cf_variation=n3lo_cf_variation,
+                                    n3lo_cf_variation=variation,
                                 ),
                             )
                         )
--- a/src/yadism/coefficient_functions/asy/kernels.py
+++ b/src/yadism/coefficient_functions/asy/kernels.py
@@ -109,6 +109,7 @@ def generate_heavy_asy(esf, nf, pto_evol, ihq):
         if not is_pv:
             n3lo_cf_variation = esf.info.theory["n3lo_cf_variation"]
             for c, channel in (("g", "Gluon"), ("s", "Singlet")):
+                variation = n3lo_cf_variation[0]
                 if c == "g":
                     variation = n3lo_cf_variation[0]
                 if c == "s":
--- a/src/yadism/coefficient_functions/asy/kernels.py
+++ b/src/yadism/coefficient_functions/asy/kernels.py
@@ -118,7 +118,7 @@ def generate_heavy_asy(esf, nf, pto_evol, ihq):
                 n3lo_g_variation = 0
                 n3lo_q_variation = 0
             for c, channel in (("g", "Gluon"), ("s", "Singlet")):
-                variation = n3lo_cf_variation[0 if c == "g" else 1]
+                variation = n3lo_g_variation if c == "g" else n3lo_q_variation
                 for res in range(pto_evol + 1):
                     name = "Asy" + ("N" * res) + "LL" + channel
                     for av in ("AA", "VV"):
--- a/src/yadism/coefficient_functions/asy/kernels.py
+++ b/src/yadism/coefficient_functions/asy/kernels.py
@@ -108,12 +108,17 @@ def generate_heavy_asy(esf, nf, pto_evol, ihq):
         )
         if not is_pv:
             n3lo_cf_variation = esf.info.theory["n3lo_cf_variation"]
+            if "F2" in kind:
+                n3lo_g_variation = n3lo_cf_variation[0]
+                n3lo_q_variation = n3lo_cf_variation[1]
+            elif "FL" in kind:
+                n3lo_g_variation = n3lo_cf_variation[2]
+                n3lo_q_variation = n3lo_cf_variation[3]
+            else:
+                n3lo_g_variation = 0
+                n3lo_q_variation = 0
             for c, channel in (("g", "Gluon"), ("s", "Singlet")):
-                variation = n3lo_cf_variation[0]
-                if c == "g":
-                    variation = n3lo_cf_variation[0]
-                if c == "s":
-                    variation = n3lo_cf_variation[1]
+                variation = n3lo_cf_variation[0 if c == "g" else 1]
                 for res in range(pto_evol + 1):
                     name = "Asy" + ("N" * res) + "LL" + channel
                     for av in ("AA", "VV"):
--- a/src/yadism/coefficient_functions/heavy/kernels.py
+++ b/src/yadism/coefficient_functions/heavy/kernels.py
@@ -70,14 +70,15 @@ def generate(esf, nf, ihq):
             ihq,
             is_pv,
         )
-        n3lo_g_variation = esf.info.theory["n3lo_cf_variation"][0]
-        n3lo_q_variation = esf.info.theory["n3lo_cf_variation"][0]
         if "F2" in kind:
             n3lo_g_variation = esf.info.theory["n3lo_cf_variation"][0]
             n3lo_q_variation = esf.info.theory["n3lo_cf_variation"][1]
-        if "FL" in kind:
+        elif "FL" in kind:
             n3lo_g_variation = esf.info.theory["n3lo_cf_variation"][2]
             n3lo_q_variation = esf.info.theory["n3lo_cf_variation"][3]
+        else:
+            n3lo_g_variation = 0
+            n3lo_q_variation = 0
         gVV = kernels.Kernel(
             weights["gVV"],
             pcs.GluonVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_g_variation),
--- a/src/yadism/coefficient_functions/heavy/kernels.py
+++ b/src/yadism/coefficient_functions/heavy/kernels.py
@@ -70,15 +70,20 @@ def generate(esf, nf, ihq):
             ihq,
             is_pv,
         )
-        n3lo_cf_variation = esf.info.theory["n3lo_cf_variation"]
+        if "F2" in kind:
+            n3lo_g_variation = esf.info.theory["n3lo_cf_variation"][0]
+            n3lo_q_variation = esf.info.theory["n3lo_cf_variation"][1]
+        if "FL" in kind:
+            n3lo_g_variation = esf.info.theory["n3lo_cf_variation"][2]
+            n3lo_q_variation = esf.info.theory["n3lo_cf_variation"][3]
         gVV = kernels.Kernel(
             weights["gVV"],
-            pcs.GluonVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation[0]),
+            pcs.GluonVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_g_variation),
         )
         gAA = kernels.Kernel(weights["gAA"], pcs.GluonAA(esf, nf, m2hq=m2hq))
         sVV = kernels.Kernel(
             weights["sVV"],
-            pcs.SingletVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation[1]),
+            pcs.SingletVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_q_variation),
         )
         sAA = kernels.Kernel(weights["sAA"], pcs.SingletAA(esf, nf, m2hq=m2hq))
         return (gVV, gAA, sVV, sAA)
--- a/src/yadism/coefficient_functions/heavy/kernels.py
+++ b/src/yadism/coefficient_functions/heavy/kernels.py
@@ -70,12 +70,15 @@ def generate(esf, nf, ihq):
             ihq,
             is_pv,
         )
+        n3lo_g_variation = esf.info.theory["n3lo_cf_variation"][0]
+        n3lo_q_variation = esf.info.theory["n3lo_cf_variation"][0]
         if "F2" in kind:
             n3lo_g_variation = esf.info.theory["n3lo_cf_variation"][0]
             n3lo_q_variation = esf.info.theory["n3lo_cf_variation"][1]
         if "FL" in kind:
             n3lo_g_variation = esf.info.theory["n3lo_cf_variation"][2]
             n3lo_q_variation = esf.info.theory["n3lo_cf_variation"][3]
+        else
         gVV = kernels.Kernel(
             weights["gVV"],
             pcs.GluonVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_g_variation),
--- a/src/yadism/coefficient_functions/heavy/kernels.py
+++ b/src/yadism/coefficient_functions/heavy/kernels.py
@@ -73,12 +73,12 @@ def generate(esf, nf, ihq):
         n3lo_cf_variation = esf.info.theory["n3lo_cf_variation"]
         gVV = kernels.Kernel(
             weights["gVV"],
-            pcs.GluonVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation["C2g"]),
+            pcs.GluonVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation[0]),
         )
         gAA = kernels.Kernel(weights["gAA"], pcs.GluonAA(esf, nf, m2hq=m2hq))
         sVV = kernels.Kernel(
             weights["sVV"],
-            pcs.SingletVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation["C2g"]),
+            pcs.SingletVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation[1]),
         )
         sAA = kernels.Kernel(weights["sAA"], pcs.SingletAA(esf, nf, m2hq=m2hq))
         return (gVV, gAA, sVV, sAA)
--- a/src/yadism/coefficient_functions/heavy/kernels.py
+++ b/src/yadism/coefficient_functions/heavy/kernels.py
@@ -73,12 +73,12 @@ def generate(esf, nf, ihq):
         n3lo_cf_variation = esf.info.theory["n3lo_cf_variation"]
         gVV = kernels.Kernel(
             weights["gVV"],
-            pcs.GluonVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation),
+            pcs.GluonVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation["C2g"]),
         )
         gAA = kernels.Kernel(weights["gAA"], pcs.GluonAA(esf, nf, m2hq=m2hq))
         sVV = kernels.Kernel(
             weights["sVV"],
-            pcs.SingletVV(esf, nf, m2hq=m2hq, n3lo_cf_variation=n3lo_cf_variation),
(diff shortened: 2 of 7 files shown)


src/yadism/runner.py
```diff
@@ -136,6 +136,14 @@ def __init__(self, theory: dict, observables: dict):
             M2target=theory["MP"] ** 2,
             fonllparts=new_theory["FONLLParts"],
             n3lo_cf_variation=theory["n3lo_cf_variation"],
+            # also accept only one number as cf variation
+            if isinstance(n3lo_cf_variation, int): 
+                n3lo_cf_variation = {
```
felixhekhorn: A dictionary might be complicated to pass around - I'd rather use a tuple together with the agreement that the first entry is `C2g` etc. (like in eko)

src/yadism/runner.py
```diff
@@ -136,6 +136,14 @@ def __init__(self, theory: dict, observables: dict):
             M2target=theory["MP"] ** 2,
             fonllparts=new_theory["FONLLParts"],
             n3lo_cf_variation=theory["n3lo_cf_variation"],
+            # also accept only one number as cf variation
```
felixhekhorn: you can not do this inside a dict of course! you must do it before (in a slightly modified way) ...

src/yadism/runner.py
jekoorn: ok. I'll try to mirror what eko does

src/yadism/runner.py
```diff
@@ -135,7 +135,9 @@ def __init__(self, theory: dict, observables: dict):
             M2W=theory["MW"] ** 2,
             M2target=theory["MP"] ** 2,
             fonllparts=new_theory["FONLLParts"],
+            # the n3lo_cf_variations are ordered as (C2g, C2q, CLg, CLq)
```
felixhekhorn: Keep the backwards compatibility you had before: if it is a single number, put that number everywhere. You just need to do this check above

src/yadism/coefficient_functions/heavy/kernels.py
felixhekhorn, line 46: `kind` tells you the structure function so "F2", "FL, "xF3", "g1" etc.

felixhekhorn: > So indeed it is ok that you don't see the distinction there

not yet - but you must depend on that of course

src/yadism/coefficient_functions/asy/kernels.py
```diff
@@ -109,6 +109,11 @@ def generate_heavy_asy(esf, nf, pto_evol, ihq):
         if not is_pv:
             n3lo_cf_variation = esf.info.theory["n3lo_cf_variation"]
             for c, channel in (("g", "Gluon"), ("s", "Singlet")):
+                variation = n3lo_cf_variation[0]
+                if c == "g":
+                    variation = n3lo_cf_variation[0]
+                if c == "s":
+                    variation = n3lo_cf_variation[1]
```
felixhekhorn: This should be practically identical to `heavy` I think

src/yadism/runner.py
felixhekhorn, line 144: are you sure we don't need some kind of casting here?

jekoorn: My idea was that with 
`theory["n3lo_cf_variation"] = (theory["n3lo_cf_variation"], theory["n3lo_cf_variation"], theory["n3lo_cf_variation"], theory["n3lo_cf_variation"])`
casting would not be necessary. Maybe I am not seeing it but what's the risk of not using casting?

src/yadism/coefficient_functions/asy/kernels.py
```diff
@@ -108,7 +108,17 @@ def generate_heavy_asy(esf, nf, pto_evol, ihq):
...
+                n3lo_g_variation = n3lo_cf_variation[0]
+                n3lo_q_variation = n3lo_cf_variation[1]
+            elif "FL" in kind:
+                n3lo_g_variation = n3lo_cf_variation[2]
+                n3lo_q_variation = n3lo_cf_variation[3]
+            else:
+                n3lo_g_variation = 0
+                n3lo_q_variation = 0
             for c, channel in (("g", "Gluon"), ("s", "Singlet")):
+                variation = n3lo_cf_variation[0 if c == "g" else 1]
```
felixhekhorn: this is not the same as for heavy. This line is too much. You must use `n3lo_g_variation` or `n3lo_q_variation` of course

tests/yadism/cf/test_nc_kernels.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, n):
         self.coupling_constants = MockCouplingConstants()
         self.m2hq = [1.0, 2.0, 3.0]
         self.threshold = Atlas(matching_scales=self.m2hq, origin=(1.65**2, 4))
-        self.theory = {"n3lo_cf_variation": 0, "pto": 3}
+        self.theory = {"n3lo_cf_variation": "0, 0, 0, 0", "pto": 3}
```
felixhekhorn: Re casting: I think I got confused by this line; i.e. this one is wrong. It must be a list a integers, but we will define it in the runcard like so - see https://github.com/NNPDF/nnpdf/blob/aa6deea48b71081eda79ea66a9fd05e3ad396f08/nnpdf_data/nnpdf_data/theory_cards/41042000.yaml#L45

felixhekhorn (changes requested):
Let me ask again: have you actually run this code? and check e.g. that when you put different numbers this is propagated down to adani correctly?

jekoorn:
> have you actually run this code?

I am currently checking this with pinefarm (since I have to generate grids anyway and needed a pipeline). But probably it's easier to do this with yadism standalone 🙈

src/yadism/coefficient_functions/asy/kernels.py
jekoorn: forgot to remove this line, thanks

jekoorn:
There are still some NaNs spawning, apparently. I will investigate this..

Radonirinaunimi:
> There are still some NaNs spawning, apparently. I will investigate this.. A fresh yadism also gets NaNs though, so I am not sure to what extent the NaNs are "supposed" to be there or not..

Could you please provide more details (which cards are you using, do the NaNs appear for specific processes/datasets/variations, etc.)? There are some NaNs that are fine (such as at very low $Q^2$) and there are some that are dangerous. From the recent grid computations, the latter should no longer appear.

jekoorn:
>Could you please provide more details (which cards are you using, do the NaNs appear for specific processes/datasets/variations, etc.)?

Sure!
For checking whether the variations produce different outputs, instead of using cards, I am using my own script that computes the structure functions. Just electron-proton collision at 1TeV. 
I am integrating $Q^2$ from as low as 2.0, and x > 1e-6. I am using the theory cards from https://github.com/NNPDF/nnpdf/pull/2494 but with the CF variations put in. 
The broader issue is here that the outputs from different variations still yield the exact same output. I am wondering now if the setup I use for checking this is actually sufficient.
See https://vp.nnpdf.science/nVd3gBnfT2S-psG9B5g5gQ==/ for the script
Probably a good check would be whether running with actual pinecards yields NaNs, but for now I wanted to check whether the CF variations do what they are supposed to.

felixhekhorn:
> The broader issue is here that the outputs from different variations still yield the exact same output.

I suggest to put a debugger (or print or whatever) here
https://github.com/NNPDF/yadism/blob/467615f723b41550d2a13c78898c9be6db0a993e/src/yadism/coefficient_functions/heavy/f2_nc.py#L49
and check what variation arrives there and then what you get back

Also do you get finite numbers with the default settings? i.e. without any variation? so we know the problem is with the variations (and not with something else)

jekoorn:
> put a debugger (or print or whatever) here
 
 nothing looks to arrive there, with whatever variation I put in. Probably explains why the calculations yield the same result between different variations. It makes me doubt whether the correlated variations were even doing what they should in the first place, before starting this implementation.
 
 With default settings I also get NaNs, but to be precise: they don't appear in the output, it's just a notification that the runner passes:
 
```
 took 6.43 s
[14:03:10] CRITICAL Some NaNs are encountered and set to zero!                                                                                                                                                         runner.py:224
           CRITICAL Some NaNs are encountered and set to zero!                                                                                                                                                         runner.py:224
           CRITICAL Some NaNs are encountered and set to zero!                                                                                                                                                         runner.py:224
           CRITICAL Some NaNs are encountered and set to zero!                                                                                                                                                         runner.py:224
           CRITICAL Some NaNs are encountered and set to zero!
```

Radonirinaunimi:
> > put a debugger (or print or whatever) here
> 
> nothing looks to arrive there, with whatever variation I put in. Probably explains why the calculations yield the same result between different variations. It makes me doubt whether the correlated variations were even doing what they should in the first place, before starting this implementation.
> 
> With default settings I also get NaNs, but to be precise: they don't appear in the output, it's just a notification that the runner passes:
> 
> ```
>  took 6.43 s
> [14:03:10] CRITICAL Some NaNs are encountered and set to zero!                                                                                                                                                         runner.py:224
>            CRITICAL Some NaNs are encountered and set to zero!                                                                                                                                                         runner.py:224
>            CRITICAL Some NaNs are encountered and set to zero!                                                                                                                                                         runner.py:224
>            CRITICAL Some NaNs are encountered and set to zero!                                                                                                                                                         runner.py:224
>            CRITICAL Some NaNs are encountered and set to zero!
> ```

These NaNs are fine (as I per above) so you shouldn't worry about those. They are in the regions where pQCD breaks down and these warnings are just there warn people that these regions are probed.

felixhekhorn:
> They are in the regions where pQCD breaks down and these warnings are just there warn people that these regions are probed.

@Radonirinaunimi I think you put these warnings there in the first place, right? so you know that they are in this region? because this can not be inferred from the output ... (i.e. if we have the kinematics in hand there it might be could to echo them alongside :upside_down_face: )

> nothing looks to arrive there, with whatever variation I put in. Probably explains why the calculations yield the same result between different variations. It makes me doubt whether the correlated variations were even doing what they should in the first place, before starting this implementation.

@jekoorn then a) you should check on master what you get there (and confirm or reject your doubts) and b) then you "just" need to trace their way back and e.g. check what you get immediately inside yadism = in the runner e.g. here https://github.com/NNPDF/yadism/blob/17a6c55cec9d1eaaf7a510a6e761c16555fc5831/src/yadism/runner.py#L127

jekoorn:
on it @felixhekhorn

Radonirinaunimi:
> @Radonirinaunimi I think you put these warnings there in the first place, right? so you know that they are in this region? because this can not be inferred from the output ... (i.e. if we have the kinematics in hand there it might be good to echo them alongside 🙃 )

Yes, I agree. I should have printed out the kinematics alongside for full transparency and certainty. Since we are at it, I would suggest to add this in this PR.

src/yadism/runner.py
```diff
@@ -235,6 +235,8 @@ def replace_nans_with_0(self, out):
                     for tup in range(2):
                         # Set any NaN or inf values in the array to 0
                         values[tup][~np.isfinite(values[tup])] = 0.0
+                        logger.critical(f"NaNs are encountered and set to zero at x = {point.x}!")
```
Radonirinaunimi: We also need to know the $Q^2$ as that's the most important bit.

jekoorn:
For debugging purposes, which FNS should I use? In some cases the code simply breaks, which probably has to do with not choosing the appropriate nf, in combination with the FNS. At least the variations look like they're activated with the FFN0 scheme. 
Maybe in the end, seeing what happens with a full pinefarm run is best after all.

felixhekhorn:
> For debugging purposes, which FNS should I use?

I suggest FFNS3, so `FNS: FFNS` + `NfFF: 3` (check that you actually pass by where you want). FFN0 would work as well and you would just end up then on the `asy` side. Then later in the real world we should also test full FONLL (which is a combination of FFNS and FFN0).

> Maybe in the end, seeing what happens with a full pinefarm run is best after all.

This would be (very likely) directly the FONLL case. However, since FONLL has multiple ingredients that might make debugging more complicated ... but actually, you could also just use the correct [FFNS theory card](https://github.com/NNPDF/nnpdf/blob/f5207f109cace9f0a0f51589126831b634ec9755/nnpdf_data/nnpdf_data/theory_cards/41002103.yaml) there (courtesy of https://github.com/NNPDF/nnpdf/pull/2498)

src/yadism/runner.py
```diff
@@ -213,9 +219,9 @@ def replace_nans_with_0(self, out):
 
         # Loop through each observable in the dictionary
         for observable, points in out2.items():
+
```
felixhekhorn:
```suggestion
```

jekoorn:
This exact [FFNS theory card](https://github.com/NNPDF/nnpdf/blob/f5207f109cace9f0a0f51589126831b634ec9755/nnpdf_data/nnpdf_data/theory_cards/41002103.yaml) with FFNS3 fails, for some reason. Also on the master branch without the `n3lo_cf_variation : [0,0,0,0]`, with the same error. See here
```
 • F2_charm at 1 pts

                                                                                                                       Calculation
yadism took off! please stay tuned ...
a heavy vv gluon is made, with variation -1
Traceback (most recent call last):
  File "/data/theorie/jkoorn/git/yad-4ihou/run.py", line 150, in <module>
    F2_tot, FL, F2_c, FL_c = runner(
                             ~~~~~~^
        Ebeam=1000,
        ^^^^^^^^^^^
    ...<6 lines>...
        theory_path=f"/data/theorie/jkoorn/git/yad-4ihou/{theory}.yaml"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/data/theorie/jkoorn/git/yad-4ihou/run.py", line 135, in runner
    F2_c = compute_observable("F2_charm", xgrid, Q2grid, Ebeam, PTO, channel, particle, target,runningscales, PDF, theory_path)
  File "/data/theorie/jkoorn/git/yad-4ihou/run.py", line 77, in compute_observable
    output = yadism.run_yadism(theory, observable)
  File "/data/theorie/jkoorn/git/yad-4ihou/src/yadism/__init__.py", line 31, in run_yadism
    return runner.get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/data/theorie/jkoorn/git/yad-4ihou/src/yadism/runner.py", line 292, in get_result
    results[idx] = elem.get_result()
                   ~~~~~~~~~~~~~~~^^
  File "/data/theorie/jkoorn/git/yad-4ihou/src/yadism/esf/esf.py", line 189, in get_result
    self.compute_local()
    ~~~~~~~~~~~~~~~~~~^^
  File "/data/theorie/jkoorn/git/yad-4ihou/src/yadism/esf/esf.py", line 143, in compute_local
    val, err = conv.convolve_vector(
               ~~~~~~~~~~~~~~~~~~~~^
        rsl, self.info.configs.managers["interpolator"], convolution_point
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/data/theorie/jkoorn/git/yad-4ihou/src/yadism/esf/conv.py", line 230, in convolve_vector
    c, e = convolution(cf, convolution_point, polynomial_f)
           ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/jkoorn/git/yad-4ihou/src/yadism/esf/conv.py", line 151, in convolution
    res, err = scipy.integrate.quad(
               ~~~~~~~~~~~~~~~~~~~~^
        quad_ker,
        ^^^^^^^^^
    ...<4 lines>...
        points=breakpoints,
        ^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/data/theorie/jkoorn/miniconda/envs/yadism-d/lib/python3.13/site-packages/scipy/integrate/_quadpack_py.py", line 479, in quad
    retval = _quad(func, a, b, args, full_output, epsabs, epsrel, limit,
                   points)
  File "/data/theorie/jkoorn/miniconda/envs/yadism-d/lib/python3.13/site-packages/scipy/integrate/_quadpack_py.py", line 639, in _quad
    return _quadpack._qagpe(func, a, b, the_points, args, full_output,
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                            epsabs, epsrel, limit)
                            ^^^^^^^^^^^^^^^^^^^^^^
TypeError: only 0-dimensional arrays can be converted to Python scalars
```

If you have an idea what is going wrong here I would be very happy to continue working on this but I think I am currently too lost by which FNS's work and which don't 😅

At least I'venow verified that the correct variations are passed to the correct places, so at least the original goal of this test has been succesful!

felixhekhorn:
ehm, I'm confused, because I have seen numbers from @evagroenendijk supposedly using that card - @evagroenendijk can you please confirm?

That being said, the error sounds more like a local problem: are you sure your runcards are 100% correct? the question is in particular for the types: are things which should be a scalar number just that and not some array? (For dumping runcards to file even `np.float` vs. Python `float` makes a difference, but I think that should not be a problem here.) However, doing remote debugging is tricky :upside_down_face: ... actually, the type mismatch might also be at the bottom (so where we are interested in).

> At least I'venow verified that the correct variations are passed to the correct places, so at least the original goal of this test has been succesful!

a good thing for sure, but you know people (meaning Juan :see_no_evil: ) want to see the real deal

jekoorn:
>However, doing remote debugging is tricky 🙃
 
 I understand, of course, but since it's apparently an issue that only pops up with some FNS's (code runs fine under ZM-VFNS), I was hoping... 
 
 But if @evagroenendijk has a runscript with this card that works, so much the better!

evagroenendijk:
@jekoorn @felixhekhorn Indeed I use exactly that theory card and it works fine. I think it's a compatibility issue between numpy and the scipy integration. I reproduce your error when using numpy v2.4, I myself was using v2.2.6 and then it does work (I think v2.4 numpy outputs an array somewhere while v2.2.6 outputs a value)

felixhekhorn:
> I think it's a compatibility issue between numpy and the scipy integration

ooh, you are saying this is a side effect of our Python (and dependencies) upgrade ... in particular #383

Okay, so we need to sort that out. I would suggest to solve this issue first (starting from `master` branch) and then rebase/merge this one on top afterwards so the two things get disentangled.

Sigh... if only yadism would have some sort of benchmarks, e.g. here #162 or also just in pinefarm, which would run on a _regular_ basis and where we would detect such problems ...

jekoorn:
> Okay, so we need to sort that out.

Will be addressed in https://github.com/NNPDF/yadism/pull/389

felixhekhorn:
> then rebase/merge this one on top afterwards

please still check explicitly that you get different numbers for different variations (and they are not crazy different)

jekoorn:
Differences are in the 10-30% range when comparing the correlated vs uncorrelated min/max variations - typically the uncorrelated variations (squares summed) are slightly smaller. I'm thinking about how to properly present the results, but I can just link a file with a plaintext table that I made using `pineappl diff`. 
Btw, all using Eva's nice benchmark runcards.

felixhekhorn:
> Differences are in the 10-30% range when comparing the correlated vs uncorrelated min/max variations

Can we be a bit more specific? for which observable? for which kinematics? we are talking only about the N3LO corrections (and not the full thing), right?

> I'm thinking about how to properly present the results, but I can just link a file with a plaintext table that I made using `pineappl diff`.

If you can come up quickly with some clever representation, fine, else also plain text files will do for now. Then for the 4.1 paper we need to do something more clever :upside_down_face:

jekoorn:
> for which observable? for which kinematics?

In this case, I used Eva's runcards for the N3LO benchmark to compute F2C and FLC. To get the N3LO corrections, should I just subtract the NNLO component? And do these observables suffice, or do we want an actual cross-section in the end?

felixhekhorn:
> In this case, I used Eva's runcards for the N3LO benchmark to compute F2C and FLC.

in principle fine, but we need to remember, that some of the kinematics are just unreachable by HERA

> And do these observables suffice, or do we want an actual cross-section in the end?

thus, it might make also sense to look at the true settings as well. We can start with the easy thing first (just F2+FL). For the paper we need the real thing.

> To get the N3LO corrections, should I just subtract the NNLO component? 

Yes. Also PineAPPL can do it directly (I think). Both things (total and relative) are interesting.

jekoorn:
https://vp.nnpdf.science/rHIQtzZPT32Fi7FTq577zA==/[F2C_FLC_all-Q2.pdf](https://vp.nnpdf.science/rHIQtzZPT32Fi7FTq577zA==/F2C_FLC_all-Q2.pdf)

plot here for the F2C and FLC SFs, compared to the NNLO predictions, the shaded band is the old correlated variation, and the colored ones are the new up/down variations for the CFs, both around the same central theory.

juanrojochacon:
Thanks for the nice plot @jekoorn can you add the IHOU (new) band, to have an idea of how it compares with the IHOU (old) band?

